PR #13, opened 2017-08-11
Allow to delete shots

frankrousseau:
There was no way to delete a shot from the web GUI. This PR aims to solve that problem. 

It fixes another bug. To delete a shot you need to delete all its tasks. It was annoying because when you delete the tasks, even if it was related to a shot, it returned to the asset list. Now, it returns to the shot or asset list depending on the entity type related to the task.

### DIFF
--- a/src/components/Shots.vue
+++ b/src/components/Shots.vue
@@ -216,7 +216,7 @@ export default {
       this.$store.dispatch('deleteShot', {
         shot: this.shotToDelete,
         callback: (err) => {
-          if (!err) this.modals.isDeleteDisplayed = false
+          if (!err) this.$router.push('/shots')
         }
       })
     },
--- a/src/components/Task.vue
+++ b/src/components/Task.vue
@@ -419,6 +419,8 @@ export default {
     confirmDeleteTask () {
       this.loading.deleteTask = true
       this.errors.deleteTask = false
+      const type = this.currentTask.entity_type_name
+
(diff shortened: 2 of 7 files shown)


mottosso:
About this, what are your thoughts on deletion overall? I can't quite tell what's happening in this change, but is it physically removing entries from the database? If so, what are your thoughts on instead of deleting, just hiding?

What I'm thinking is that deleting anything is dangerous due to references and relationships. Because we're working with files of various formats, it can be difficult to know whether a file is referencing and depends something from the database. Deleting anything could cause unexpected behaviour, whereas hiding would preserve existing references, but prevent further ones from being made.

In any system I develop, I actively prohibit deletion of anything in the database because of this.

frankrousseau:
Thank you @mottosso for your feedback about this feature. Currently when the user runs a deletion of a shot, Kitsu asks to the API to remove the shot from the database. From the user perspective, it should disappear from the listing. We could imagine a trash system, but that would probably be more confusing than useful. 

So your concerns are more related to Zou, the API. About Zou currently the full deletion occurs as long as there is no data linked to it (no foreign key pointing to the shot entry in the database). So there are  few chances something related to the shot exists when the deletion occurs. 

Ideally, a deleted flag or a deletion date, would be safer. But I don't want to rush to add this feature, it will add some complexity with which I don't want to deal with.

mottosso:
Thanks @frankrousseau, that makes sense.

PR #106, opened 2018-07-05
Create DEVELOPMENT.md

mottosso:
Hi @frankrousseau,

I've been trying to reduce the number of steps involved in getting a development environment up for Kitsu, and here's what I've come up with. I expect it can get shorter still, which is why I'm sharing it here. There are a few things left that I was hoping I could get your take on.

1. I'm utilising the `cgwire/cgwire` Docker container for Zou and friends, which is running on a dedicated VM on Windows, so what I've done is adjusted the `config.js` to reflect where the database is. On Linux, Docker would be running natively, so it wouldn't need any change. But I expect for OSX Docker also runs in a VM with a dedicated IP.
2. Live Updates; because Docker doesn't carry the filesystem events, editing files locally doesn't trigger a rebuild in CGWire which is kind of a pain. I've found that a workaround is to (1) enter the docker image and (2) "touch" the file manually. This would cause `npm run` to detect the change and trigger a rebuild for that file. Any idea about how this could be made more smooth?
3. Builds take over a minute to complete; this may be normal, but if not is there anything that could be done to speed that up?
4. There are a number of dependencies added on-top of the vanilla `npm install`, I was wondering whether these could be included somehow? I'm not too familiar enough with the npm dependency mechanism to know how this works.

Let me know what you think!

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,143 @@
+### Kitsu development environment
+
+The [original documentation](https://kitsu.cg-wire.com/development-environment/) for setting up Kitsu includes setting up a significant amount of tooling on the local system - e.g. node.js, 900+ (!) dependencies, postgres, redis, FFMPEG and more.
+
+The below method involves just `git` and `docker`.
+
+1. Clone Kitsu
+1. Edit to support Dockerised Zou
+1. Create Docker volume
+1. Run Docker build environment
+
+<br>
+
+### Usage
+
+We'll clone Kitsu and use it to override the Kitsu from the already Dockerised distribution of CGWire.
+
+```bash
+docker run -d --rm -p 80:80 --name cgwire cgwire/cgwire
+git clone https://github.com/cgwire/kitsu.git
+cd kitsu
+```
+
+From here we need to modify the setup script of Kitsu, to use Zou via Docker container instead of a local version.
+
+<details>
+  <summary>On Windows?</summary>
+<br>
+If you are using Docker Toolbox on Windows, save this to <code>kitsu.patch</code> and run <code>git am < kitsu.patch</code>.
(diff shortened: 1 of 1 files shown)


frankrousseau:
That's a good idea. But I think it can be done in a simpler way:

https://kitsu.cg-wire.com/development-environment/

I will see how to rewrite your tutorial next week

mottosso:
Sorry I should have been more clear; this wasn't ready for merging yet, I was hoping to resolve the above issues first. Nonetheless, when you've got something let me know!

mottosso:
For the override in index.js, is there any way I can control the `proxyTable` addresses via environment variables?

Specifically, [this](https://github.com/cgwire/kitsu/blob/6e81c667971f750a905f3138df0f0d9d05076308/config/index.js#L32) and [this](https://github.com/cgwire/kitsu/blob/6e81c667971f750a905f3138df0f0d9d05076308/config/index.js#L39).

To use the `cgwire/cgwire` Docker image, we need those to point to different addresses.

- `http://127.0.0.1:5000` -> `http://192.168.99.100/api`
- `http://127.0.0.1:5001` -> `http://192.168.99.100/socket.io`

frankrousseau:
We could probably change the config file by using environment variable:

`process.env.API_TARGET || http://127.0.0.1:5000`

frankrousseau:
The Dockerfile should only be used to have an up and running API. The development environment must be build outside the Docker container via a git clone. From the cloned repositiory you can run `npm install` and `npm run dev`.

mottosso:
> We could probably change the config file by using environment variable:

Giving this a go.

> The Dockerfile should only be used to have an up and running API. 

Yes, was just having a look at that, and it looks to work very well.

PR #130, opened 2018-09-13
Add fps, ratio and resolution at the project scale

frankrousseau:
**Problem**

* Last comment cell doesn't give all information
* Avatar thumbnail is sometimes broken
* There is no way to tell the CG artist what are the expected FPS, ratio and resolution of a project.

**Solution**

* Add a tooltip to the last comment cell with the full comment
* Fix condition to display avatar thumbnails (person)
* Allow to set fps, ratio and resolution on a project in the production list
* Display the project info as a tooltip of project avatar in the CG artist todo list

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -115,6 +115,11 @@ td.actions {
   min-width: 100px;
 }
 
+ul {
+  list-style-type: disc;
+  margin-left: 1em;
+}
+
 .hero {
   background-color: #CFCFCF;
 }
--- a/src/components/Topbar.vue
+++ b/src/components/Topbar.vue
@@ -28,7 +28,7 @@
             <div class="level-item">
               <combobox
                 class="production-selector"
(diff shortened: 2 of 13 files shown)


tokejepsen:
Would it be possible to set other metadata than just FPS, ratio and resolution?

Maybe like an optional JSON dictionary with arbitrary metadata.

frankrousseau:
Hello @tokejepsen, thank you for this idea. it's feasible but we are working on other stuff (especially TV show support). I can include this feature request when we will allow to add metadata to assets and shots.

I don't know if we already talk about it but we also provide consultancy services to add features. It means we can change our priorities if you are ready to pay us to do the work. We already did that with 4 other studios.

tokejepsen:
Great! Is there an issue tracker for feature requests?

frankrousseau:
It's the same as the issue list of this repo. We will tag it as enhancement.

PR #193, opened 2019-02-11
Fix typo for English no_assignation_right

BigRoy:
Fix typo "no allowed" to "not allowed" + Also made first character uppercase.

**Problem**
The English sentence for not being allowed to assign tasks was a bit odd with a typo.

**Solution**
Fix typo "no allowed" to "not allowed" + Also made first character uppercase.


frankrousseau:
Cool! Thank you @BigRoy!

PR #196, opened 2019-02-13
Rotate Delete Search Filter button to show as "X"

BigRoy:
Correctly rotate the "delete search filter query-list" button so it looks like an X instead of a +. Also increased size slightly so it's relatively similar size as before it was rotated.

**Problem**
The remove search query list filter shows as a Plus (+) instead of a Cross (X).

**Solution**
Change the CSS so it's a rotated plus.

See image below:

search_query_list_delete_rotate

---

Note: I only tested the CSS in my browser and have **not** tested this on a live Kitsu server. I also don't know if this CSS is used elsewhere in the theme.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -631,7 +631,7 @@ input.search-input:focus {
 
 .query-list .tag .delete {
   margin-left: 0.5em;
-  transform: scale(0.6)
+  transform: rotate(45deg) scale(0.7)
 }
 
 .query-list .tag:hover {


BigRoy:
To elaborate some more:

Please note that the [bulma.css already rotates the `.delete` classes](https://github.com/BigRoy/kitsu/blob/master/public/css/bulma.css#L2830) but because the `App.vue` file was overwriting the `transform` with a smaller scale it was overriding the original transform as declared in bulma.css.

frankrousseau:
Thank you for this very clear PR and for the fix. Your solution is totally fine. Thank you a lot!

PR #218, opened 2019-03-18
Allow jpg and jpeg in default filter for Change Avatar/Thumbnail

BigRoy:
**Problem**

The Production change avatar/thumbnail default filter doesn't allow .jpg or .jpeg files even though when you change the filtering to all and upload a jpg it works just fine.

**Solution**

Set the default filter to include `.jpg` and `.jpeg`.

---

#### Notes

_Note: Code is untested. I justed searched for the `.png` in the code and refactored those that seemed related. **Please test first**_

Don't know if [this line](https://github.com/cgwire/kitsu/blob/4f91b507a879e9656e45cfbe856d127c5ea9e2af/src/store/modules/tasks.js#L511) also needs updating but I don't think so, since I believe this refers to the converted file after upload. Right?


frankrousseau:
Great thank you!

PR #319, opened 2019-11-26
Spinner test

GoOz:
Add tests for Spinner component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "babel-loader": "8.0.5",
     "core-js": "2.6.9",
     "eslint": "5.16.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "babel-loader": "8.0.5",
     "core-js": "2.6.9",
     "eslint": "5.16.0",
+    "eslint-plugin-jest": "^23.0.4",
```
frankrousseau: Please can you remove the `^`? It brings unexpected behaviour.

GoOz: Oh right! Good call.

PR #336, opened 2019-12-06
Add tests for RowActions component

GoOz:
**Problem**
`RowActions` component has no tests

**Solution**
`RowActions` component has **now** tests

### DIFF
--- a/src/components/widgets/RowActions.vue
+++ b/src/components/widgets/RowActions.vue
@@ -2,6 +2,7 @@
 <td class="actions has-text-right">
   <router-link
     class="button"
+    data-test="button-edit"
     :to="editRoute"
     v-if="!hideEdit && !entry.canceled"
   >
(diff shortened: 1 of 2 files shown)


frankrousseau:
Are you sure it's a good idea to add information to the component that are use only for tests?

GoOz:
> Are you sure it's a good idea to add information to the component that are used only for tests?

I know some people do that to ensure that even if the DOM around the element changes in any way (order or CSS classes, etc) the tests doesn't break for nothing. CF http://blog.rstankov.com/using-rel-in-testing/ for example.

Also, in this case, I have issues finding the right elements and differentiate them from one another without going from `shallowMount` to `Mount` and test either the route or the loaded icon.

frankrousseau:
Ok. Have a look at references too. In vue-js they are useful to select an element.

PR #342, opened 2019-12-16
Add Test for shotline component

GoOz:
**Problem**
`ShotLine` component has no tests

**Solution**
`ShotLine` component has **now** tests


frankrousseau:
Thx!

PR #344, opened 2019-12-18
add color picker for writing annotations

GoOz:
**Problem**
Annotations can be difficult to see if the preview image has the same or a close color as the one used for annotations.

**Solution**
Added the possibility to change the color of annotations.


frankrousseau:
Awesome, thank you!

PR #346, opened 2019-12-23
Enhanced annotations reloaded

GoOz:
**Problem**
The new color picker for annotations was only available in the `PlaylistPlayer`.

**Solution**
There is now a tested `ColorPicker` component, added to `PlaylistPlayer`, `VideoPlayer` and `PicturePlayer`.


frankrousseau:
Great, thank you!

PR #347, opened 2019-12-24
Fix: Always show Add Tasks Button

### LINKED ISSUE #345: "Add tasks" button is hidden
On the global page of the shots or asset, the Add Task button is hidden.
You have to hover the first line (the Name line) to see the Add Task button appears and be able to click on it.

GoOz:
**Problem**
Add Tasks button was only displayed on hover

**Solution**
Add Tasks button is now always displayed.

Fix #345

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -707,7 +707,7 @@ tr .actions p {
   margin-bottom: 0;
 }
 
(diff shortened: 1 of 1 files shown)


frankrousseau:
Thx!

PR #355, opened 2020-01-23
Add CSVtoJSON lib to display data in new Preview Modal

GoOz:
**Problem**
The user can't see what a CSV look like before importing.

**Solution**
Added a preview modal before importing

### DIFF
--- a/src/components/pages/Assets.vue
+++ b/src/components/pages/Assets.vue
@@ -143,14 +143,14 @@
     @confirm="confirmDeleteMetadata"
   />
 
-  <preview-modal
-    :active="modals.isPreviewDisplayed"
+  <import-render-modal
+    :active="modals.isImportRenderDisplayed"
     :is-loading="loading.importing"
     :is-error="errors.importing"
-    :parsedCSV="parsedCSV"
+    :parsed-csv="parsedCSV"
     :form-data="assetsCsvFormData"
-    @reupload="reuploadFile"
-    @cancel="hidePreviewModal"
+    @reupload="resetImport"
+    @cancel="hideImportRenderModal"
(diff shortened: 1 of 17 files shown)


src/components/pages/Assets.vue
```diff
@@ -143,14 +143,26 @@
     @confirm="confirmDeleteMetadata"
   />
 
+  <preview-modal
+    :active="modals.isPreviewDisplayed"
+    :is-loading="loading.importing"
+    :is-error="errors.importing"
+    :parsedCSV="parsedCSV"
```
frankrousseau: Can you use only snake case for parameters?

src/components/pages/Breakdown.vue
```diff
@@ -427,6 +451,39 @@ export default {
...
+      this.processCSVFile(file)
+        .then((results) => {
+          this.parsedCSV = results
+          this.hideImportModal()
+          this.loading.importing = false
+          this.showPreviewModal()
+        })
+    },
+
+    reuploadFile () {
```
frankrousseau: The function name doesn't seem to match the behavior. It looks more like a reset function.

src/components/pages/Shots.vue
```diff
@@ -705,7 +710,7 @@ export default {
       this.uploadShotFile((err) => {
         if (!err) {
           this.loading.importing = false
-          this.hideImportModal()
+          this.hidePreviewModal()
```
frankrousseau: Preview is a term use everywhere else to describe videos and pictures sent by artists. Can you look for another term like ImportRender?

frankrousseau:
Thx!

PR #362, opened 2020-02-04
Add Selectboxes in CSV render headers to switch columns

GoOz:
**Problem**
If a user made a mistake in the name of a column in its imported CSV, the CSV must be updated then reuploaded.

**Solution**
The user can switch a wrong column header to one of the required columns header.

### DIFF
--- a/src/components/modals/ImportRenderModal.vue
+++ b/src/components/modals/ImportRenderModal.vue
@@ -47,10 +47,10 @@
             <col
               v-for="(cell, index) in parsedCsv[0]"
               :key="`col-${index}`"
-              :class="stateColumn(cell, columns)"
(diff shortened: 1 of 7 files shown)


src/components/modals/ImportRenderModal.vue
```diff
@@ -174,25 +186,56 @@ export default {
...
+    columnOptions () {
+      let options = [{
+        label: this.$t('main.csv.choose'),
+        value: this.$t('main.csv.unknown')
+      }]
+      this.columns.forEach(item => {
+        options.push({ label: item, value: item })
+      })
+      return options
+    },
```
frankrousseau: Can you separate all functions with a blank line?

PR #383, opened 2020-02-18
Import update

GoOz:
**Problem**
When a user wants to import data via CSV, the import render modal doesn't differentiate new data from old one (already in the database) and just ignore those when importing.

**Solution**
In the import render modal, the user can now :
- see the data that will be ignored on import because it already exists in the database
- see the data that will update the already existing one in the database if the option "Update" is toggled.


src/components/pages/People.vue
```diff
@@ -203,6 +209,17 @@ export default {
...
       }
+    },
+
+    filteredPeople () {
+      const persons = []
+      this.displayedPeople.forEach(people => {
+        let person = ''
+        person += people.first_name
+        person += people.last_name
+        persons.push(person)
```
frankrousseau: You should use a map. Relying on `includes` is too slow. It leads to a O(n2) complexity.

```
const personKey = `${people.first_name}-${people.last_name}`
persons[personKey] = true
```

PR #386, opened 2020-02-19
Allow to change all the task types of a playlist

frankrousseau:
**Problem**

It is not possible to change all the task types of a playlist at the same time. Which can be annoying for long playlists when they are melted.

**Solution**

Add a button and a modal to allow to chose a task type and apply it to the whole playlist.


src/components/modals/SelectTaskTypeModal.vue
```diff
@@ -0,0 +1,114 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+.is-danger {
+  color: #ff3860;
+  font-style: italic;
+}
+
+.mt4 {
```
GoOz: It's not used anywhere in the file, is this typo?

frankrousseau: I fixed it. Thx

PR #388, opened 2020-02-19
Fix/tests

GoOz:
**Problem**
After updating **Vue-test-utils** from *1.0.0-beta.29* to *1.0.0-beta.31*, several tests were failing.

**Solution**
Vue-test-utils *1.0.0-beta.30* removed sync updates and forced async updates therefore a lot of tests failed.
Fixed this by turn sync updates to async ones.

### DIFF
--- a/tests/unit/pages/production-schedule.spec.js
+++ b/tests/unit/pages/production-schedule.spec.js
@@ -1,10 +1,10 @@
+import Vue from 'Vue'
 import { shallowMount, createLocalVue } from '@vue/test-utils'
 import vuescroll from 'vue-scroll'
 import Vuex from 'vuex'
 import moment from 'moment'
 
 import i18n from '../../../src/lib/i18n'
-import { range } from '../../../src/lib/time'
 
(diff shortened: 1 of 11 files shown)


tests/unit/pages/production-schedule.spec.js
```diff
@@ -145,22 +142,25 @@ describe('ProductionSchedule', () => {
     test('onBodyScroll', () => {
     })
 
-    test('convertScheduleItems', () => {
-      let item = wrapper.vm.convertScheduleItems([{
+    test('convertScheduleItems', done => {
+      const item = wrapper.vm.convertScheduleItems(undefined, [{
```
GoOz: Not sure about the `undefined` (that refers to `taskTypeElement`) but that's the only way I found to make it work. Any thought ?

frankrousseau: This probably requires to add tests to test both cases. But that's not the matter of the PR. So let's merge this.

Issue #377, opened 2020-02-14
The first artist added to a team isn't displayed in the team list


gwendupre:
When you create a new production (short, tv-show, or feature film), and add people on the production team's page, the first person added won't appear on the list until you refresh the page, or go to another page and come back.

PR #395, opened 2020-02-25
Fix: first team members added are displayed

GoOz:
fix: #377

**Problem**
When you add persons to a production's team for the first time `this.currentProduction.team` is updated but its update does not trigger the `teamPersons` computed property. It should but it doesn't.

**Solution**
`state.currentProduction.team` being undefined for new production it can't be reactive. Declaring an empty value for a new production's team makes it reactive.


frankrousseau:
I think you should add a websocket listener for when a team change occurs on another computer.

PR #404, opened 2020-03-05
Add Drag & Drop priority for TaskTypes

GoOz:
**Problem**
In order to change the priority of a task type, the user must edit each task type manually.

**Solution**
Added a drag & drop functionnnality that allow the user to quickly change the order of task types and therefore their own priority level.

### DIFF
--- a/src/components/lists/TaskTypeList.vue
+++ b/src/components/lists/TaskTypeList.vue
@@ -22,13 +22,19 @@
 
   <div class="table-body" v-scroll="onBodyScroll">
     <table class="table splitted-table">
-      <tbody>
-        <tr class="type-header">
+      <draggable
+        v-model="assetsItems"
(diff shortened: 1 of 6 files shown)


src/components/lists/TaskTypeList.vue
```diff
@@ -130,6 +148,47 @@ export default {
...
+    },
+
+    updatePriorityShots () {
+      this.shotsItems.forEach((item, index) => {
+        index += 1
+        const form = {
+          id: item.id,
+          name: item.name,
+          priority: String(index),
+          for_shots: String(item.for_shots || 'true'),
```
GoOz: For some reason that I don't totally get, if I don't specify `for_shots` & `allow_timelog` here the existing values change from `true` to `false`…

Issue #403, opened 2020-03-03
Daily quota don't display


gwendupre:
In Short, TV show, Feature film production.

The daily quotas of the artist don't display in the current week, even if the average column is displaying the right data.

![image](https://user-images.githubusercontent.com/42962633/75782529-5191b000-5d5f-11ea-9062-a52ff345e90d.png)

PR #408, opened 2020-03-09
Fix: quota day display when the day date is 1 digit

GoOz:
fix: #403

**Problem**
When displaying quotas by day, quotas in the first 9 days didn't display because it was 1 digit for the day date when the data sent was 2 digits for the day date, so it didn't match.

**Solution**
Enforcing date format of 2 digits for days as well.

### DIFF
--- a/src/components/pages/quota/Quota.vue
+++ b/src/components/pages/quota/Quota.vue
@@ -360,16 +360,20 @@ export default {
     },
     monthToString,
 
+    dateDigit (date) {
+      return date.toString().padStart(2, '0')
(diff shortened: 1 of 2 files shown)


frankrousseau:
Great thank you!

frankrousseau:
Can you add tests for this?

GoOz:
> Can you add tests for this?

For the whole page or just for the method ?

frankrousseau:
For this method

frankrousseau:
Thx!

PR #412, opened 2020-03-11
feat: Add Metadata option hidden from client

GoOz:
**Problem**
A user can't hide a metadata column from a client

**Solution**
Added an option "hidden from client" in metadata modal creation/edition


frankrousseau:
Merged manually

PR #414, opened 2020-03-19
Feat/new tables

GoOz:
**Problem**
The actual data tables used in Shots page had several layout issues.

**Solution**
Refactoring of the table, aimed to use as much as possible native HTML/CSS behaviour on tables.
It should replace all data tables in the near future but for now, in order to spot potential new issues, it's used in the Shots page only.

### DIFF
--- a/src/components/lists/ShotList.vue
+++ b/src/components/lists/ShotList.vue
@@ -463,6 +463,12 @@ export default {
       this.displayMoreShots()
     },
 
+    // Remaining function for retrocompatibility
+    resizeHeaders () {
+      return true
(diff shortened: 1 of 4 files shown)


src/components/lists/ShotList.vue
```diff
@@ -401,6 +416,27 @@ export default {
       )
     },
 
+    visibleColumns () {
```
GoOz: Not sure if it's the best way to count all the columns in a table but that's all I could think of. I'm open to suggestions to enhance this part if there is any.

PR #417, opened 2020-03-23
Fix/new table

GoOz:
**Problem**
- Thumbnails with images are not well positionned
- Long words should break when there's not enough room in a cell

**Solution**
- Added some missing css for thumbnails with images
- Add ellipsis on metadata content


frankrousseau:
Thx!

PR #420, opened 2020-03-24
feat: Add resizable columns feature

GoOz:
**Problem**
Users can't resize columns at will.

**Solution**
Add a feature to allow users to manually resize each columns independantly.


frankrousseau:
Great! But can you persist the chosen size in the local storage?

GoOz:
oh right, good call.

PR #421, opened 2020-03-25
feat: Save resized column's width in localStorage

GoOz:
**Problem**
When a user changes the size of a column, this size is lost when refreshing or changing page.

**Solution**
Added custom sizes in the user localStorage


frankrousseau:
Can you allow only metadata columns to be resizable?

PR #423, opened 2020-03-26
feat: new Table for Assets vue

GoOz:
**Problem**
Assets page don't have new table layout and column resizing

**Solution**
Assets page now have new table layout and column resizing


frankrousseau:
Great, thank you!

PR #435, opened 2020-04-07
Various fixes

frankrousseau:
**Problem**

* When creating an asset in all section in TV show, it fails
* It is not possible to edit comment status

**Solution**

* Use route dedicated to asset creation instead of generic one used for entities.
* Add a task status combobox to edit comment form.

### DIFF
--- a/src/locales/fr.js
+++ b/src/locales/fr.js
@@ -69,6 +69,7 @@ export default {
     revision: 'révision',
     set_status_to: 'Positionner le statut à',
     unpin: 'Désépingler',
+    text: 'Text',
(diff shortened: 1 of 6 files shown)


src/locales/fr.js
```diff
@@ -69,6 +69,7 @@ export default {
     revision: 'révision',
     set_status_to: 'Positionner le statut à',
     unpin: 'Désépingler',
+    text: 'Text',
```
GoOz, line 72:
```suggestion
    text: 'Texte',
```

frankrousseau: thx

PR #456, opened 2020-04-29
fix: path stroke ratio on videoplayer for annotation

GoOz:
**Problem**
When a user draws an annotation path on smaller videoplayer then draw another one on the bigger player, stroke ratio is not respected.

**Solution**
Correction of the `strokeMutliplier` var with the right data, which would be `annnotation.width` instead of `obj.canvasWidth`


frankrousseau:
Great, thank you!

PR #473, opened 2020-06-16
Added toggle mute button to VideoPlayer.

ricekab:
**Problem**
There is currently no way to toggle sound in VideoPlayer.

**Solution**
This adds a sound / mute toggle button between the repeat button and time indication bar. The icons used are volume-2 and volume-x . Implementation is based on the way play/pause is implemented.


frankrousseau:
Awesome, thank you!

PR #542, opened 2020-10-06
Feature/russian locale

piiq:
**Problem**
Russian language was not supported in Kitsu.

**Solution**
Added Russian localisation. In this implementation for most of the terms the professional slang is used. 
For example:
- the word "asset" is translated as "ассет" (pronounced as asset) because everyone calls assets this way.
- same goes for the word "shot". It is translated as "шот" (pronounced as shot) because this is how everyone calls shots.

There might be a more old-school way of calling production terms from the soviet cinematographic legacy, but there's nobody I know who uses such terminology.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,21 @@ To install the development environment, follow [the dedicated guide](https://kit
 
 ## Sponsors
 
-[![Les Fées Spéciales](https://www.cg-wire.com/images/logo-les-fees-speciales.png | width=100)](http://les-fees-speciales.coop)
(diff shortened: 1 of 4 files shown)


frankrousseau:
Awesome! Thank you a lot @piiq and @NikolaiGavchuck! 

For the maintenance, could you use POEditor once I uploaded your work?
 https://poeditor.com/join/project?hash=fpUejpWDVo

piiq:
@frankrousseau sure!

PR #584, opened 2020-11-29
Add support to Opera (browser) > 22

HuelderAntunes:
**Problem**
Browser validation in routes.js does not include Opera. Opera is a popular browser based on chromium, with all the basic features to work with Kitsu.

**Solution**
Add support from version 22 on the isValidBrowser configuration object inside routes file.


frankrousseau:
Thank you @HuelderAntunes for your contribution!

PR #619, opened 2021-01-19
Add brief for production

Jermie:
- Add a component ProductionBrief,
- Add ProductionBrief to tab on ProductionSettings,
- Add locales en
- Add description in method `editProduction`

**Problem**

The first refresh of `brief` is not working without a direct call to `setProduction`

### DIFF
--- a/src/components/pages/ProductionSettings.vue
+++ b/src/components/pages/ProductionSettings.vue
@@ -3,6 +3,11 @@
     <div class="wrapper">
     <div class="tabs">
       <ul>
+        <li :class="{'is-active': isActiveTab('brief')}">
+          <a @click="activeTab = 'brief'">
+            {{ $t('brief.title')}}
+          </a>
+        </li>
         <li :class="{'is-active': isActiveTab('taskStatus')}">
           <a @click="activeTab = 'taskStatus'">
             {{ $t('task_status.title')}}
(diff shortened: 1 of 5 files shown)


src/components/pages/production/ProductionBrief.vue
```diff
@@ -0,0 +1,106 @@
...
+
+    async editBrief () {
+      this.isLoading = true
+      await this.editProduction(
+        {
+          ...this.value,
+          ...{ description: this.brief }
+        }
+      )
+      if (this.currentProduction.id === this.value.id) {
```
Jermie, line 83: I had to put this "ugly" things, this is needed for the first refresh

src/components/pages/ProductionSettings.vue
```diff
@@ -21,6 +26,10 @@
       </ul>
     </div>
 
+    <div class="tab" v-show="isActiveTab('brief')">
+      <ProductionBrief :value="currentProduction" />
```
frankrousseau, line 30: The value is not needed because the current production is available through mapGetters function.

src/components/pages/production/ProductionBrief.vue
```diff
@@ -0,0 +1,106 @@
...
+      return (!str || str.length === 0)
+    },
+
+    openEditing () {
+      this.isEditing = true
+    },
+
+    async editBrief () {
+      this.isLoading = true
+      await this.editProduction(
```
frankrousseau, line 77: Errors are not handled here.

PR #621, opened 2021-01-20
Add production parameters tab on production settings page

Jermie:
- Add component production-parameters and somes tests
- Add locales
- Update DateField to update the local value if the props "value" change
- Integrate production-parameters in production settings page

**Problem**

The edition of the production settings is not really accessible, some fields likes the start date, end date, number of episodes and span between episodes are missing in the current form.

**Solution**

Add a tab in production setting page to allow users to modify all fields in a new form


frankrousseau:
Awesome, thank you! @Jermie

PR #622, opened 2021-01-21
Add somes columns for production task status to display more data

Jermie:
**Problem**
Production task status doesn't display some interesting data like,  `is_done`, `is_retake` `is_artist_allowed` and `is_client_allowed` on settings page.

**Solution**
Add column to the existing settings page to display the data.


frankrousseau:
Great, thank you!

PR #626, opened 2021-01-26
some fixes

cbsrobot:
just some minor fixes that I saw while playing around.


frankrousseau:
That will for sure fix many glitches! Thank you!

PR #629, opened 2021-01-30
minor fixes

cbsrobot:
here's the screenshot for the change in Logs.vue : https://imgur.com/a/XswpKIK
not sure why no person was logged for this event.


frankrousseau:
Thank you for this PR. I don't understand the commit "return to shots". First, you use a method while a computed property would be better. Second, you replace the computed property `shotsPath` which does already what you try to achieve. Can you give me more information?

frankrousseau:
I was able to reproduce your issue. Let me a bit of time to fix it properly and I will integrate your changes.

PR #639, opened 2021-02-19
Annotation

cbsrobot:
**Problem**
Clicking or stopping the Player won't show the annotations.


frankrousseau:
Thank you for your help!

PR #659, opened 2021-03-23
people: fix departments tests

HugoDelval:
**Problem**

2 tests were failing in departments.spec.js

**Solution**

- the code is now using async/await + try/catch instead of then().catch(), it's easier to test this way.
- refacto of the test file which was not respecting the eslint.js config


frankrousseau:
For your information, I don't like async/await, it's less "functional" and it doesn't match the rest of the code.

PR #663, opened 2021-03-24
[assets/shots]: ability to select a column

HugoDelval:
**Problem**

When selecting tasks we don't want to scroll to select the whole column

**Solution**

Add a button on the column header which selects the whole column

image

Depends on https://github.com/cgwire/kitsu/pull/667


src/components/mixins/entity_list.js
```diff
@@ -192,6 +193,27 @@ export const entityListMixin = {
       this.showHeaderMenu()
     },
 
+    onSelectColumn () {
+      const mapOfElements = this.assetMap ? this.assetMap : this.shotMap
+      const listOfElements = Object.values(mapOfElements)
+      const selection = []
+      for (let i = 0; i < Object.keys(mapOfElements).length; i++) {
```
frankrousseau, line 200: In JS, it's better to use `forEach` or `map` functions. For loops are not commonly used.

PR #664, opened 2021-03-24
[assets/shots] hide also time_spent/fps/estimation

HugoDelval:
**Problem**

We want to have more control over the columns displayed in the assets and shots list

**Solution**

We can now show/hide those columns : 

- fps
- time spent
- frames in
- frames out
- estimation

Depends on https://github.com/cgwire/kitsu/pull/667


frankrousseau:
Thank you!

PR #671, opened 2021-03-26
[people] fix excluding search + search builder

HugoDelval:
**Problem**

- The search was ignoring the `-` prefix (to perform a exclusion)
- We couldn't build a search graphically

**Solution**

- fix the search to include the `-` check
- add modal to build the search dynamically, more specifically the departments


frankrousseau:
Thanks!

PR #704, opened 2021-05-20
Select multiple lines a time

HugoDelval:
**Problem**
- A feature was wanted: the ability to select multiple assets/shots at a time
- small issue when selecting multiple tasks

**Solution**
the issue is fixed and..

feature added:
- select a first asset/shot
- hold the shift key
- select another asset/shot
- all assets/shots in-between are now selected


frankrousseau:
Great! thank you!

PR #714, opened 2021-06-09
[production] quick create productions objects

HugoDelval:
**Problem**
It's long to create the first objects when creating a production

**Solution**
This PR adds a page on which we can add several standard thing a production needs to have to be fully configured

It still lacks assets and shots

image

-----

image


HugoDelval:
This last commit adds the ability to import shots/assets : 

image

PR #735, opened 2021-08-24
[playlist] show pictures a given number of frames

HugoDelval:
**Problem**
The pictures werent shown when launching a playlist

**Solution**
The pictures are now shown a given and customable amount of time (48 frames per default) before going to the next picture/video

image


frankrousseau:
It looks great. But can you set by default the shot `nb_frames`. If it's empty can you set 48?

HugoDelval:
So this means the number of frames a pictures shows on screen is different from picture to picture?

HugoDelval:
I've done the modification Frank, tell me if that's what you had in mind

PR #744, opened 2021-09-15
[playlist] add play button on images

HugoDelval:
**Problem**
We couldn't start the playlist when we were on a picture

**Solution**
This PR adds a play/pause button on images, starting the playlist

image


frankrousseau:
Great, thank you!

PR #745, opened 2021-09-16
Various fixes

frankrousseau:
**Problem**

* In the add comment form, pasting clears the preview field.
* Comparison syncing in the side video player is still broken.
* Annotations are broken in extra wide mode in the side preview player.

**Solution**

* Apply file pasting only if the attachment and the preview modals are active.
* Don't use smart functions to sync videos but sync them with raw functions.
* Resize properly the annotaion canvas each time the extra wide mode is activated.


EvanBldy:
Thank you @frankrousseau

PR #748, opened 2021-09-16
Change default selection in combobox

LedruRollin:
**Problem**
When a user wants to add tasks, he arrives on this modal :

image

image
The default value of the combobox is set on "For project". However, it may lead to mistakes since a distracted user might think he's creating task for his current shot page, when he'll actually fill all his project with tasks.

**Solution**
A simple fix is to swap the default combobox values : if someone mistakenly creates tasks for current list instead of whole project, the user simply has to change the setting and launch again, preventing him from having to delete a lot of unwanted tasks.

I didn't know anything about Vue.js before this PR, so if I missed something in this minor change, let me know

### DIFF
--- a/src/components/modals/CreateTasksModal.vue
+++ b/src/components/modals/CreateTasksModal.vue
@@ -122,7 +122,7 @@ export default {
       form: {
         task_type_id: ''
       },
-      selectionOnly: 'false',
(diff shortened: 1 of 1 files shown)


frankrousseau:
It seems good, thank you @LedruRollin for your contribution!

PR #757, opened 2021-09-22
[playlist] fix play next image

HugoDelval:
**Problem**
When there were several images in a row in a playlist, the playlist was stopping at each one of them. The issue was not visible when there was a mix of videos/images and was not spotted before.

**Solution**
Force the playlist to continue playing when the next plan is an image


frankrousseau:
Ok thank you!

PR #782, opened 2021-11-16
Fix webpack problem + lint every js files

EvanBldy:
**Problem**

- Webpack v5 not supported along vue-cli v4, we have to downgrade to webpack v4 (a vue-cli v5 which support webpack v5 is in RC). https://next.cli.vuejs.org/migrations/migrate-from-v4.html
- Some files are not linted

**Solution**

- Downgrade webpack to latest v4
- npm run lint on every files


frankrousseau:
Great, thank you!

PR #798, opened 2022-01-14
Make task type tag readable in dark theme

ex5:
**Problem**
Text on the task type tag is almost unreadable in dark theme, see, for example, the Task Types page:

2022-01-14-153302_1028x938_scrot

**Solution**
Overriding text color under `.dark` class in TaskTypeName fixes the issue:

2022-01-14-153319_1017x932_scrot


frankrousseau:
Thank you @ex5 for this helpful PR. I had already fixed the problem in my codebase. I have to close this PR. Sorry for that.

PR #800, opened 2022-01-20
Mattermost notifications with Webhook/Matterhook

Ramel:
**Problem**
- Send Kitsu notifications to Mattermost (Tested on Mattermost Version: 5.37.2)

**Solution**
- Need also to update Zou to same build;
- Perhaps a rebase error with "src/locales/en.js" file, which was modified after. Line 699 notifications_slack_user: 'Slack member ID',

Mattermost setup
=================
Need to create in Mattermost a User, for example **"Kitsu"**, as **"Team Admin"** (so it can set a Webhook).  
Need to create a Channel for the notifications, for example **"Kitsu"**.
Connect in Mattermost with this user account **"Kitsu"**. Go to **"Integrations"** / **"Incoming Webhooks"** / **"Add incoming Webhook"** with :
- Title: Kitsu
- Channel: Kitsu (created before)
- Username: Kitsu

Then copy the URL generated to Kistu web interface, in **Organisation** settings view, Integrations and fill **"Mattermost Webhook (Optional)"** fields with it.

Then, each user can set in his profile, the **"Mattermost notifications Enabled"** field to "Yes" and enter his **"Mattermost Username"** to receive Kitsu Notifications into Mattermost.


frankrousseau:
Hello @Ramel,

Thank you for this contributiion. Is it ok for us to merge this PR?

Ramel:
As it's my first Pull request here, I hope it's ok!
My only concern is how the Flask migration is done in final delivery as we have added columns in the db.
But the code for me si ok and tested our own server!

.eslintrc.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
     'jest/no-focused-tests': 'error',
     'jest/no-identical-title': 'error',
     'jest/prefer-to-have-length': 'warn',
-    'jest/valid-expect': 'error'
+    'jest/valid-expect': 'error',
+    'indent': 'off'
```
frankrousseau, line 21: Why did you remove this? Indentation should follow a strict scheme.

frankrousseau:
Thank you a lot for this contribution! It will improve the life of other studios for sure!

.eslintrc.js
Ramel: We only have added the `'indent': 'off'`, if not set the compilation hangs.
But it's perhaps a thing to keep in mind! 
Perhaps specific to our Dev config? Don't known if we should have keep it in the final commit.

frankrousseau: I removed it. The compilation hangs because it waits that you fix the indentation of the mentioned files. It's here to make sure the code stays similar everywhere :).

PR #820, opened 2022-03-09
Playlist-/PreviewPlayer: allow backspace to delete annotations

cbsrobot:
**Problem**
osx maps the <delete> key to <backspace> which makes deleting annotations cumbersome. if you want press the <delete> key you need to press the <fn> modifier too.

**Solution**
allow the delete and the backspace key to delete annotations.


EvanBldy:
Hello @cbsrobot,
Thanks a lot for your PR :) 
Have a nice day

PR #821, opened 2022-03-09
vue.config.js: use 127.0.0.1 as api proxy address instead of localhost

cbsrobot:
On osx the proxy cannot be forwarded  from http://127.0.0.1:8080
to http://localhost:5000 . By changing the address to an ip address
it also matches the socket.io address below.


EvanBldy:
Hello @cbsrobot,
Thanks a lot for the PR :)

PR #825, opened 2022-03-14
Minor fixes to Edits page

ex5:
**Problem**
Clipboard - March 14, 2022 4 51 PM

Should be this instead:
2022-03-14-170450_363x67_scrot

**Solution**
Updated `en.js` and added a missing bracket in `EditList.vue`.


EvanBldy:
Hi @ex5,
Thanks for a lot for your PR :)
Can edits stay plural ? 
Because we can have multiple edits.

PR #846, opened 2022-04-11
[departments] in department filter for entity view and add-metadata-m…

EvanBldy:
…odal filter departments by entity type

**Problem**
In entities views (Shots, Assets, Edits) there's a problem with the combobox filter for departments (it can have departments not related to the entity type)

**Solution**
In the combobox filter for departments only allow to select departments related to the entity type (filter them by the entity type in the task type)


frankrousseau:
Thx!

PR #847, opened 2022-04-11
Feature: Several episodes for an asset

Tilix4:
**Problem**
As a supervisor, I need to know in which episodes my asset will be used. I need to set this information when preparing the production for the artists, even when no shot has been created already.

**Solution**
- `Breakdown` casting creation: added an `All` setting to list all episodes and being able to assign assets to them.
  *When selecting an episode in the combobox, the behaviour becomes the same as it is currently*
 image
- In `Assets` panel, the episodes the asset is casted in are listed
 image
- Added events for `update:` in Breakdown and Assets.

*NB: I left some `TODO` to raise some concerns and keep some questions I need invoice for in mind.*

See implementation: https://github.com/cgwire/zou/pull/452


frankrousseau:
The code was added to the main branch.

PR #868, opened 2022-05-03
Feature: task types for asset types

Tilix4:
**Problem**
For some type of assets, mainly in 2D productions, the task types differ. An environment is not made the same way as a character.
I'd like to be able to design a task types workflow for an asset type.

**Solution**
- In `asset types`:
  - possibility to set task types during creation/edition (by default all existing task types are added to the asset type), using the handy system used in Production creation: 
    image
  - task types are displayed in list:
    image
- In `Assets`, task types columns which are not used for an asset type are dashed and unselectable. Task are not created for these task types for relevant asset types:
  image

#### Implementation
https://github.com/cgwire/zou/pull/461


Tilix4:
- [ ] Add forgotten`widget/AssetTypeName.vue` to git repository

frankrousseau:
Code squashed and added to the main branch.

Issue #880, opened 2022-05-25
Links in the Schedule section appear to be broken


ex5:
The task type links on the left side of the Schedule page lead to broken task type schedules.

These are the links in question:

![2022-05-25-112131_1193x1318_scrot](https://user-images.githubusercontent.com/176934/170229150-f2a351f1-387c-4b5f-9b0c-afa30bf38cdd.png)

This is what is displayed when they are clicked (the page is empty, although it should have schedule for the Rigging task type):

![2022-05-25-112153_1195x590_scrot](https://user-images.githubusercontent.com/176934/170229195-b28c04ba-c9f9-45c1-bb8a-eb71e9ecf022.png)

It looks like the same task type schedules can be reached via the links in the header of the tasks table:

https://user-images.githubusercontent.com/176934/170228581-d211cf09-1e1e-4fe2-969f-d573f6cba996.mp4

which probably means that it's the way routes are constructed at the Schedule page is broken. Looking at the routes, the working one has a plural (`assets`):

broken route:
`.../asset/task-types/f6294e61-de1c-4287-aa17-9dc0815a93fb/schedule`

working route:
`.../assets/task-types/f6294e61-de1c-4287-aa17-9dc0815a93fb/schedule`

`ProductionSchedule.vue` is using `this.$tc(taskType.for_entity.toLowerCase(), 2)` which keeps returning a singular instead of a plural, so maybe that's the reason, not sure what would be a good fix for this, would appreciate some help with that.

EvanBldy:
Hi @ex5,
Thanks for the report
I have fixed the problem in this PR https://github.com/cgwire/kitsu/pull/882 can you try it and close the issue if the problem is fixed for you ? 
Have a nice day

ex5:
yep, the schedule links all work on latest `master` :+1: 
thank you @EvanBldy !

PR #882, opened 2022-05-27
[qa] fix links in schedule when we click on task types

EvanBldy:
**Problem**
https://github.com/cgwire/kitsu/issues/880
Links in the schedule page are broken

**Solution**
Fix link construction (assets, shots, edits needs to be plural)


frankrousseau:
Thx!

PR #885, opened 2022-05-31
[comments] export csv of comments of a task

EvanBldy:
**Problem**
We can't export comments of a task in a csv.

**Solution**
Add a button in TaskInfo to export comments of a task in a csv.


src/components/sides/TaskInfo.vue
```diff
@@ -267,6 +273,9 @@ import Spinner from '@/components/widgets/Spinner'
 import SubscribeButton from '@/components/widgets/SubscribeButton'
 import TaskTypeName from '@/components/widgets/TaskTypeName'
 import PreviewPlayer from '@/components/previews/PreviewPlayer'
+import csv from '../../lib/csv'
```
frankrousseau, line 276: Can you use '@' import? And place the lib import before the widget imports?

EvanBldy: I have changed all relatives imports with '@' imports

PR #906, opened 2022-06-28
[qa] Fix extract annotations when player is in fullscreen

EvanBldy:
**Problem**
- when play is in fullscreen we can't extract annotations

**Solution**
- fix extract annotations when play is in fullscreen


frankrousseau:
Thx!

PR #913, opened 2022-07-07
Various fixes

frankrousseau:
**Problem**

* Shot CSV export headers are broken.
* Annotations are not saved anymore when browsing tasks.
* @frame tagging is not accurate.
* Clients should not see comment authors
* Annotations are not updated in real-time in the preview player

**Solution**

* Add resolution and max retakes headers to the headers when needed.U
* Use the right task ID when saving annotations
* Clean the way @frame tagging is computed (remove useless frame additions)
* Hide comment authors from the client when the author is not a client.


frame:
@frame tagging is indeed not accurate. :)

PR #958, opened 2022-10-06
[ux] Full redesign of the action bar

frankrousseau:
**Problem**

The action bar concerns only a few section, is visually very present and most of all blocks the navigation.

**Solution**

Peek 04-10-2022 12-10

* Make it floatable
* Make it smaller
* Make it cleaner: the menu is on top and allows to select sections via icons
* Non related: Add a section to mass-set thumbnails


frankrousseau:
I merged it directly on my computer.

PR #961, opened 2022-10-17
various fixes + authentication improvements

EvanBldy:
**Problem**
- In edits page department preselection is not working.
- When we create a new user the "invite" button can be clicked if user does not exist.
- In People page when a user is created the new user is not in the list if we don't reload the page.
- src/components/modals/EditPersonDepartmentsModal.vue file is unused.
- Admin can't change passwords of other users.
- In People page when we have unactive users and we scroll the column name is not staying at the left as for active users.
- When LDAP is activated among other things we can't : create new users / edit existing users (not generated from LDAP).
- We need a security when a lot of logins fails for the same user.

**Solution**
- Fix department preselection in edits page.
- When creating a new user "invite" button is now disabled if user does not exist.
- Properly reload users in People page if a new user is created.
- Remove unused file : src/components/modals/EditPersonDepartmentsModal.vue.
- Admin can now change password of other users (only can't change passwords of other admins).
- Fix scrolling issue for unactive users in People page.
- Allow to create and edit users not from ldap (and change passwords).
- When login fails multiple times for same user, force to wait before an other attempt.

This PR is related to this PR from Zou : https://github.com/cgwire/zou/pull/533


src/locales/en.js
```diff
@@ -296,6 +296,7 @@ export default {
     forgot_password: 'Forgot password?',
     login: 'Log in',
     login_failed: 'Log in failed, please verify your credentials',
+    too_much_login_failed_attemps: 'Too much login failed attemps, please wait one minute',
```
jhoolmans, line 299: Small correction:
'Too many failed login attempts, please wait one minute'

EvanBldy: Yeah thanks :)

src/components/modals/ChangePasswordModal.vue
```diff
@@ -0,0 +1,169 @@
...
+import { modalMixin } from '@/components/modals/base_modal'
+
+import TextField from '@/components/widgets/TextField'
+
+import auth from '@/lib/auth'
+
+export default {
+  name: 'change-password-modal',
+  mixins: [modalMixin],
+  props: [
```
frankrousseau, line 83: Please use descriptors for props like in the other components.

src/components/modals/ChangePasswordModal.vue
```diff
@@ -0,0 +1,169 @@
...
+  },
+
+  methods: {
+    ...mapActions([
+      'changePasswordPerson'
+    ]),
+
+    confirmClicked () {
+      if (auth.isPasswordValid(this.form.password, this.form.password2)) {
+        this.changePasswordPerson({
```
frankrousseau, line 116: Can you convert this to a promise?

src/components/pages/Login.vue
```diff
@@ -12,6 +12,7 @@
               <input
                 class="input is-medium email"
                 type="text"
+                v-model="email"
```
frankrousseau, line 15: Always put  `v-model` at the bottom for consistency reasons.

src/locales/en.js
```diff
@@ -845,7 +847,7 @@ export default {
       button: 'Change password',
       error: 'An error occurred while changing password. Please verify your current password.',
       success: 'Your password was successfully changed!',
-      unvalid: 'Your new password confirmation doesn\'t match or your password is too short (7 chars, at least, is expected).'
+      unvalid: 'Your new password confirmation doesn\'t match or your password is too short (8 chars, at least, is expected).'
```
frankrousseau, line 850: Don't add translations, they are managed via POEditor.

src/components/modals/ChangePasswordModal.vue
EvanBldy: It's fixed

src/components/pages/Login.vue
EvanBldy: It's fixed

src/components/modals/ChangePasswordModal.vue
EvanBldy: It's fixed

PR #971, opened 2022-11-09
[qa][shots] fix calculation of nb_frames

EvanBldy:
**Problem**
- The calculation of nb_frames for shots is missing one frame and is not calculated sometimes due to wrong conditions

**Solution**
- Fix calculation of nb_frames


frankrousseau:
Thank you!

PR #978, opened 2022-11-25
feat: all sequences option for breakdown episode

kaamaurice:
**Main concept**
Added an "all Sequences" option for the Breakdown per episode in sequence combobox.
- this option can display all the casting of each shots of an episode to make it easier.


frankrousseau:
Hello @kaamaurice,
Thank you for your contribution. Can you handle the "all sequences" case with something else than a number? Usually, IDs are made of strings, it can do things confusing.

frankrousseau:
Thank you!

PR #982, opened 2022-12-07
Add support for TOTP + OTP via email

EvanBldy:
**Problem**
- we don't support 2FA like TOTP or verification code via email

**Solution**
- add support for TOTP and verification code via email


src/components/pages/Login.vue
```diff
@@ -37,6 +44,25 @@
...
           </div>
+          <div
+            class="field mt2"
+            v-show="isMissingOTP || isWrongOTP"
+          >
+            <p class="control has-icon">
+              <input
+                class="input is-medium password"
+                type="text"
+                v-model="otp"
```
frankrousseau: Please, put `v-model` at the end of the attributes.

src/components/modals/ChangePasswordModal.vue
```diff
@@ -127,6 +151,13 @@ export default {
         })
     },
 
+    disableTwoFactorAuthenticationClicked () {
+      this.isErrorDisableTwoFactorAuthentication = false
+      this.isError = false
+      this.disableTwoFactorAuthenticationPerson(this.person)
```
frankrousseau, line 157: Adding loading indicators, would make things clearer to the user.

frankrousseau (commented):
It seems, ok I made some comments to improve a few more things.

PR #1034, opened 2023-03-10
Format source code with ESLint + Prettier

NicoPennec:
**Problem**
The source code is not harmonized and may contain auto-detectable errors

**Solution**
Set ESLint to fix code issues and Prettier to format the source code


src/components/cells/DescriptionCell.vue
```diff
@@ -79,25 +72,23 @@ export default {
   },
 
   computed: {
-    ...mapGetters([
-    ]),
+    ...mapGetters([]),
 
-    content () {
+    content() {
```
frankrousseau, line 77: This rule is not good. It's not coherent with the rest and I'm not sure it works.

src/components/cells/DescriptionCell.vue
```diff
@@ -153,7 +144,7 @@ td {
 
 .description-cell .tooltip {
   background-color: $white;
-  border-radius: .5rem;
+  border-radius: 0.5rem;
```
frankrousseau, line 147: Note sure about this one either.

src/components/cells/DescriptionCell.vue
NicoPennec: This rule is forced by Prettier 2.x, no options are available to disable this one, but the JS code is still valid.

src/components/cells/DescriptionCell.vue
NicoPennec: This rule is forced by Prettier too. The zero will be removed by the build process anyway.

NicoPennec:
Too much git conflicts... I created a new cleaner PR here #1038

PR #1038, opened 2023-03-13
Format source code with ESLint + Prettier

NicoPennec:
**Problem**
The source code is not harmonized and may contain auto-detectable errors.

**Solution**
Set ESLint to fix code issues and Prettier to format the source code.

This PR includes some folder exclusions (eg. locales)


frankrousseau:
Thank you and congrats for your first PR!

PR #1046, opened 2023-03-16
[i18n] Fix loading of russian language

NicoPennec:
**Problem**
From the User Profile page, the Russian language is not applied after selecting it in the form.

**Solution**
Fix the loading of the json locale file


frankrousseau:
Thank you!

Issue #1041, opened 2023-03-15
On episode page we can't select the whole task type column from the column head menu


gwendupre:
**Context**

Studio name: Gwen
Kitsu version: 0.16.3
Production type: 

**Describe the bug**

On a TVshow, when we are on the episode page, and if we want to select a whole task type column, If we use the menu from the task type header "Select column", nothing happens.

**Screenshots**

![image](https://user-images.githubusercontent.com/42962633/225281527-2bb2241b-900e-4787-a657-ba4e70819a6d.png)

**Desktop (please complete the following information):**

 - OS: Windows
 - Browser Chrome
 - Version 111.0.5563.65 (Build officiel) (64 bits)

PR #1047, opened 2023-03-17
[ux] fix column selection in lists

NicoPennec:
**Problem**
For some list components, the column selection from the header menu doesn't work:
- edit list
- sequence list
- episode list

Related to #1041 

Capture d’écran 2023-03-17 à 15 48 19

**Solution**
Pass the entities to be selected directly to the selector, for a separation concern between the mixin and the data stores


src/components/lists/ShotList.vue
```diff
@@ -9,7 +9,7 @@
         @minimize-clicked="onMinimizeColumnToggled()"
         @delete-all-clicked="onDeleteAllTasksClicked()"
         @sort-by-clicked="onSortByTaskTypeClicked()"
-        @select-column="onSelectColumn"
+        @select-column="onSelectColumn(displayedShots.flat())"
```
frankrousseau, line 12: Displayed shots don't represent all the shots. The data from the cache are needed for assets and shots.

NicoPennec: Correct, I forgot the data pagination.

PR #1049, opened 2023-03-21
Fix several lint errors

NicoPennec:
**Problem**
When running `lint` script from `package.json`, ESLint returns many errors (duplicate key, invalid code, ...)

**Solution**
Fix or refactor code to remove lint errors to improve code quality.


frankrousseau:
Thank you!

PR #1054, opened 2023-03-27
[translations] Update Russian translation

morevnaproject:
**Problem**
Some strings in UI are not shown when Russian language enabled.

screenshot_003

screenshot_004

**Solution**
This PR translates some non-translated strings.
The translation is not complete yet, but it is already better than before. :)
I will continue if everything is alright.

### DIFF
--- a/src/locales/ru.json
+++ b/src/locales/ru.json
@@ -29,8 +29,8 @@
                 "ready_for": "Готово для"
             },
             "cancel_text": "Вы действительно хотите отправить в архив {name}?",
-            "delete_for_selection": "",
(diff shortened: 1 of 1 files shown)


frankrousseau:
Hello @morevnaproject,

Thank you for your contribution.
But, Please use POEditor for translation contributions. We cannot use Github for this. You will find the link here:
https://www.cg-wire.com/en/community.html

morevnaproject:
Ah, got it! I was suspecting there should be some online translation platform, but didn't found at first attempt. Will contribute there,  thank you!

PR #1056, opened 2023-03-30
[breakdown] show an error when there's an while error saving casting

EvanBldy:
**Problem**
- There's no error shown when there's an error while saving casting.

**Solution**
- Show an error when there's an error while saving casting.

### DIFF
--- a/src/components/pages/Breakdown.vue
+++ b/src/components/pages/Breakdown.vue
@@ -161,6 +161,7 @@
               :metadata-display-headers="metadataDisplayHeaders"
               :big-mode="isBigMode"
               :is-description="isDescription"
+              :is-save-error="saveErrors[entity.id]"
               @edit-label="onEditLabelClicked"
               @add-one="addOneAsset"
               @remove-one="removeOneAsset"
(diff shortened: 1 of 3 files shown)


src/components/pages/Breakdown.vue
```diff
@@ -1095,7 +1114,15 @@ export default {
           entityId,
           casting: castingToPaste
         })
-        this.saveCasting(entityId).then(this.setLock).catch(console.error)
+        this.saveCasting(entityId)
+          .then(() => {
+            delete this.saveErrors[entityId]
```
NicoPennec, line 1119: Is it intentional to delete the error only after the promise resolution, rather than when calling the promise as below?

eg.
```js
          delete this.saveErrors[entityId]
          this.saveCasting(entityId)
            .then(() => {
              this.setLock()
            })
```

EvanBldy: Yes, it's in case you save again

EvanBldy: I have changed that

PR #1057, opened 2023-03-30
[metadata] support new types (number, boolean)

NicoPennec:
**Problem**

-  Some users need to have a numeric value in theirs custom column metadata.   
Currently it's limited to a free string value or a list of values.

**Solution**
- Extends metadata with new data types:
  - string
  - **number**
  - **boolean**
  - list
  - checklist


frankrousseau:
Great thank you!

PR #1066, opened 2023-04-12
[metadata] fix data refresh on update column name

NicoPennec:
**Problem**
After renaming a custom metadata column, the existing data on the entity list are reset to empty.

**Solution**
Fix data reactivity when updating metadata on stores.


frankrousseau:
Thank you!

PR #1068, opened 2023-04-14
Fix/input number

NicoPennec:
**Problem**
- the number input on metadata fields allows to input non decimal value.

**Solution**
- show error on invalid fields and block the metadata update.


frankrousseau:
Great, thank you!

PR #1077, opened 2023-04-21
doc(DEVELOPMENT): Typo in Node.js minimal version

pcharmoille:
**Problem**
Regarding the engine key in [package.json file](https://github.com/cgwire/kitsu/blob/master/package.json#L88) I think it was supposed `16` and not `6`

**Solution**
6 to 16


frankrousseau:
Great, thank you!

PR #1078, opened 2023-04-21
[shots] add the possibility to import an EDL

EvanBldy:
**Problem**
- There's no possibility to import an EDL file to entre frames / frame_in / frame_out

**Solution**
- Add that possibility


src/store/modules/shots.js
```diff
@@ -535,6 +539,23 @@ const actions = {
       })
   },
 
+  uploadEdlFile({ commit, state, rootGetters }, { nomenclature, match_case }) {
+    const production = rootGetters.currentProduction
+    let episode = rootGetters.isTVShow ? rootGetters.currentEpisode : null
```
NicoPennec, line 544: prefer "const"

PR #1085, opened 2023-05-03
Fix button rendering on Edit page

NicoPennec:
**Problem**
On the Edit page, the "Join review room" button is too dark on white theme.

issue-edit-button

**Solution**
Use css variables to set the white/dark theme.


frankrousseau:
Thx!

PR #1086, opened 2023-05-03
Fix/tasks

NicoPennec:
**Problem**
- When selecting / unselecting tasks (with click + ctrl), the selected list may be incorrect.

issue-selected-tasks

**Solution**
- Fix data store reactivity
- Add missing translation keys


frankrousseau:
Thank you!

Issue #1090, opened 2023-05-09
TextArea wrongly resized when editing comments


Maxiriton:
**Context**

Studio name: 3.0 Studio
Kitsu version: 0.16.14
Production type: Feature Film

When editing comments, the size of the text area automatically changes when mouse move. 

We have quite long pinned comments for each task, where our director can give his guidelines. So we edit them quite often and try to resize the TextArea to give us more space, but it seems that there is some CSS/JS rule that force the whole pop up to fit in screen, leading to the text area being forced resized. 

 - OS: macOS 13.0.1
 - Browser Firefox
 - Version 112

frankrousseau:
Thank you for the report. We are checking what is going wrong. The browser should allow you to resize it.

NicoPennec:
Hi @Maxiriton, it will be fixed in the next release.

NicoPennec:
Issue fixed in the current release (Kitsu > [0.17.x](https://github.com/cgwire/kitsu/releases/tag/v0.17.0))
Let me know if it's good for you now 😉

PR #1095, opened 2023-05-11
Improve ActionPanel component

NicoPennec:
**Problem**
- Legacy code from ActionPanel, related to its floating mode, can generate a side effect on textarea resizing with Firefox (see #1090).
- Missing translations.
- After reopening the ActionPanel, the last action is no longer selected.

**Solution**
- Remove legacy code related to floating panel.
- Add missing i18n keys.
- Fix to restore last selected action from LocalStorage.


frankrousseau:
Thank you!

PR #1096, opened 2023-05-12
[tasks] fix the width persistence of the task info panel on right side

NicoPennec:
**Problem**
- The width position of the TaskInfo panel on right side is not saved and restore after select/unselect a task or a page reload.

**Solution**
- Check the validity of the parent panel and apply the saved width.


frankrousseau:
[quantic-avocats.com](mailto:fxlanglais@quantic-avocats.com)

frankrousseau:
Thank you!

PR #1097, opened 2023-05-13
Adapt search to indexer migration changes

frankrousseau:
**Problem**

The backend switched to another indexer. The matched terms are now returned as list of strings (list of matching terms).

**Solution**

Adapt the match terms UI to this change.

### DIFF
--- a/src/components/pages/EntitySearch.vue
+++ b/src/components/pages/EntitySearch.vue
@@ -208,7 +208,7 @@ import EntityPreview from '@/components/widgets/EntityPreview'
 // import PeopleAvatar from '@/components/widgets/PeopleAvatar'
 import Spinner from '@/components/widgets/Spinner'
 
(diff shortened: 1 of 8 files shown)


src/components/pages/EntitySearch.vue
```diff
@@ -373,19 +373,12 @@ export default {
...
-      } else {
-        return this.$t('search.match_details', {
-          term,
-          target
-        })
-      }
+      const target = entity.matched_terms.join(', ')
+      const term = this.searchQuery
+      return this.$t('search.match_details', {
+        term,
```
NicoPennec, line 379: `term` it no longer exists in that updated translation: `match_details: 'match found in {target}'`

frankrousseau: Thx!

Issue #1102, opened 2023-05-30
Unable to save annotations.


henrybennettfr:
Kitsu version: 0.16.15
Production type: VFX

Ever time I draw annotations, I get this error: "An error occured while saving your annotation, please wait 3s for another try."  If I'm in fullscreen mode, this also forces the full screen mode to exit.

The annotations appear on the video correctly, but once I refresh the page, they are gone.

 - OS: Windows Server 2019 Datacenter
 - Browser Chrome Version 113.0.5672.128 (Official Build) (64-bit) 
 
 I get the same bug using Mac OS 13.3.1 (a) running Chrome 113.0.5672.126 (Official Build) (x86_64)

Thanks for the help!  It sound similar to this bug from a few years back ( https://github.com/cgwire/kitsu/issues/769 ). I'm not sure how to check which error occurs on the server side.

NicoPennec:
Hi @henrybennettfr,
It seems an error from your backend side. 
Do you know which version of Zou is installed?

In addition, can you open the "devtools" panel in your Chrome browser ? (press Command+Option+i  on MacOS)
Then go to your "network" tab and try again to create an annotation please. You will see a line "update-annotations" with more details about the backend call.

NicoPennec:
Thanks @henrybennettfr  for the screenshot, it's very useful!

About your Zou backend, you can find the version on this url: **https://`your_hostname`/api/status**

NicoPennec:
yes, an artist should not be able to add annotation.
It's an issue in the UI. We will fix it asap.
Thank you @henrybennettfr for your contribution 😉

NicoPennec:
Issue fixed since the release >= 0.17.2 of Kitsu

PR #1103, opened 2023-06-01
Fix/update annotation

NicoPennec:
**Problem**
- On the Task panel on side bar, an user with the "artist" role has access to the annotations tool, but he should not be able to. (related to #1102)

**Solution**
- Harmonize the permission check between the Task page and Task side bar, for the PreviewPlayer component.


frankrousseau:
Thanks!

PR #1104, opened 2023-06-06
[metadata] fix permission check for supervisor in departments

NicoPennec:
**Problem**
- As supervisor, I cannot update the metadata configured for my departments.

**Solution**
- Fix permission check in entity lists and breakdown page, due to missing getters.


frankrousseau:
Thx!

PR #1108, opened 2023-06-12
Various improvements

frankrousseau:
**Problem**

* It's not possible to filter search results by production, it may clutter the results.
* The team schedule doesn't show the main status bars

**Solution**

* Add a combobox to select the production to search on and send the parameter to the API.
* Get start and end date of all tasks for persons from the API. Then display them as root bars of the schedule.

### DIFF
--- a/src/store/modules/people.js
+++ b/src/store/modules/people.js
@@ -340,7 +340,9 @@ const actions = {
   },
 
   async fetchPersonTasks({ commit, state, rootGetters }, personId) {
-    const tasks = await peopleApi.getPersonTasks(personId)
+    let tasks = await peopleApi.getPersonTasks(personId)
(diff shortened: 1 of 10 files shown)


src/store/modules/people.js
```diff
@@ -340,7 +340,9 @@ const actions = {
   },
 
   async fetchPersonTasks({ commit, state, rootGetters }, personId) {
-    const tasks = await peopleApi.getPersonTasks(personId)
+    let tasks = await peopleApi.getPersonTasks(personId)
+    const doneTasks = await peopleApi.getPersonDoneTasks(personId)
+    tasks = tasks.concat(doneTasks)
```
NicoPennec, line 345: You can call these 2 endpoints in parallel with a Promise.all() for perf purposes.

PR #1119, opened 2023-06-26
[qa] fix search icon position on input of entity search page

NicoPennec:
**Problem**
Wrong css icon on Entity Search page
icon

**Solution**
Fix icon position


frankrousseau:
Thx!

PR #1120, opened 2023-06-28
Various improvements

frankrousseau:
**Problem**

* There are some typos in the locales
* There is no way to filter tasks on priorities.
* The task panel can be too wide.
* The annotations don't display properly in the newsfeed

**Solution**

* Check typos with Grammarly
* Add a filter in the task type page and in the entities build a filter modal
* Add a 900px limit to the task panel size.
* Add a settable id for the annotation canvas of the preview player. Set a different id for each player used in the newsfeed.


src/components/sides/TaskInfo.vue
```diff
@@ -1194,7 +1194,8 @@ export default {
 
     onExtendMove(event) {
       const diff = this.lastWidthX - this.getClientX(event)
-      const panelWidth = Math.max(this.lastWidth + diff, DEFAULT_PANEL_WIDTH)
+      let panelWidth = Math.max(this.lastWidth + diff, DEFAULT_PANEL_WIDTH)
+      if (panelWidth > 800) panelWidth = 900
```
NicoPennec, line 1198: error in condition?
`800` instead of `900`

PR #1135, opened 2023-07-18
Improve Tasks page

NicoPennec:
**Problem**
on Tasks page:
- resizing task on Schedule is confusing with estimation value
- estimated days with decimal values are incorrectly saved and printed in schedule, if a dot or comma separator is used
- the data table have weird style on Estimation tab
- wrong auto-resizing of parent item in Schedule, after a child resizing

**Solution**
- disable task resizing in schedule tab
- use a "number" input field and fix the estimation formatting
- fix style of estimation table
- fix max date calculation that set size of parent item


frankrousseau:
Thx!

PR #1149, opened 2023-08-16
Various improvements

frankrousseau:
**Problem**

* It's not possible to use emoji in the comment markdown.
* When there are several uploads, the upload indicators are hidden to quickly
* We cannot use sub previews to set thumbnails
* The last displayed sub preview index is not kept.
* The zoom pan mode in the preview player is not accessible clearly.
* It's not possible to insert an element in the checklist
* It's not possible to mark a checklist item with a frame
* 2 time on 3 not possible to reorder sub previews
* It would be great to have thumbnails with the revision number
* It's not possible to scale / rotate a group of annotations

**Solution**

Get Emojis unicode and use marked-emoji to display them.

+ Fix rendering issues with blockquote in dark mode
+ Fix rendering issues with list position

* Wait for all uploads to finish before showing the comment.
* Use the current sub preview to set the thumbnail.
* Keep in memory the last displayed sub preview index.
* Add a loupe icon for setting the zoom pan mode
* Send an insert-item event when enter is typed
* Add a frame marker at the beginning of each item when posting the comment
* Clean the sub preview reordering process: send event and reset previews data when the event is sent
* Show thumbnails alongside revision numbers
* Allow to scale / rotate a group of annotations


src/App.vue
```diff
@@ -1817,8 +1817,24 @@ tbody:last-child .empty-line:last-child {
   padding: 0;
 }
 
-.content ul {
-  margin: 0;
+.content {
+
+  .content ul {
```
NicoPennec, line 1822: `.content .content` is for nested "content" elements.
I'm not sure this case exists in the Comment component.

PR #1150, opened 2023-08-16
[qa][comments] fix reply comments date staying in utc+0 when displaye…

EvanBldy:
…d + prettier

**Problem**
- Reply comments date stay in utc+0
**Solution**
- Show them in the user's timezone


frankrousseau:
Thank you!

PR #1170, opened 2023-09-06
[ux] add a new combobox-tag component

NicoPennec:
**Problem**
- Users need to select multiple values from a list of choices.

**Solution**
- Create a new "combobox-tag" component to select multiple values from a list.

Capture d’écran 2023-09-06 à 12 12 06


frankrousseau:
Thank you Nicolas

Issue #1022, opened 2023-02-23
Type changing when modifying a value in frame_in and frame_out


Elsombrerobot:
**Context**

When changing the value of frame_in or frame_out on a shot in kitsu.

Studio name: Malilart
Kitsu version: 0.15.49
Production type: tv show

**Describe the bug**

When the field frame_in or frame_out is modified (on a shot here) by hand on kitsu, it changes its type to str, while it was registered as int.

**Desktop (please complete the following information):**

 - OS: Windows 10
 - Browser Chrome
 - Version 109

NicoPennec:
Hi @Elsombrerobot,
Thanks for your report!

I'm trying to understand how your data can be numeric, because from UI it's always saved as a string value.

How did you initialize your frame in/out data ? from Kitsu UI or from Gazu/Zou ?

PR #1176, opened 2023-09-11
[metadata] save number metadata as a numeric instead of a string

NicoPennec:
**Problem**
- Number metadata can be initialized as numeric value in Zou (via API or Gazu), but Kitsu then modifies it as a string (related to #1022).

**Solution**
- Always save Number metadata as numeric value on Kitsu.


frankrousseau:
Thank you!

PR #1179, opened 2023-09-11
Various improvements

frankrousseau:
**Problem**

* The validation header edit button disappears when the header is too big.
* Using the Enter key for saving a search is not intuitive and annoying. There is no need for a shortcut for this.
* The Realtime breaks filter in the task type page.

**Solution**

* Use the right CSS positioning
* Remove the enter key event listener in search fields
* Use the right task list for indexation

### DIFF
--- a/src/components/pages/TaskType.vue
+++ b/src/components/pages/TaskType.vue
@@ -504,6 +504,7 @@ export default {
   },
 
   mounted() {
+    this.searchField.setValue(this.$route.query.search || '')
     this.clearSelectedTasks()
     const isAssets = this.$route.path.includes('assets')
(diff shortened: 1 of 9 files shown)


src/components/pages/TaskType.vue
```diff
@@ -1480,8 +1498,10 @@ export default {
         ) {
           setTimeout(() => {
             this.resetTaskIndex()
+            console.log('ok')
             this.$nextTick(() => {
               if (!this.selectedTasks.get(eventData.task_id)) {
+                console.log('ok 2', this.searchField.getValue())
```
NicoPennec, line 1504: some `console.log` in this PR

PR #1183, opened 2023-09-13
fix search in feature film

NicoPennec:
**Problem**
- On the Shots and Edits pages, the search filter doesn't work if you are on a Feature Film type production.

**Solution**
- Process search on press enter.


frankrousseau:
Thx

PR #1204, opened 2023-09-27
Improve avatar loading

NicoPennec:
**Problem**
User avatar images are loaded too many times because they are not effectively cached.
This unnecessarily increases the network load.

**Solution**
- Refactor how to cache images and sync people in the UI, based on each person's last update date.
- Replace v-lazy library by native lazy loading.
- Add an option to disable lazy loading when necessary.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
         destructuring: 'all'
       }
     ],
+    'vue/component-name-in-template-casing': ['error', 'kebab-case'],
     'vue/multi-word-component-names': 'off',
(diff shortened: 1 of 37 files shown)


PeteDraper:
The images are also too big for the icon sizes - you can save out the image by right-clicking the icon to find out how big it is... for an icon it's exceptionally large. Needs downsizing for different uses akin to how the shot thumbnails work. This will also speed up page load effectively.

NicoPennec:
Hi @PeteDraper, you are completely right! This is a point that we have also identified and will look to improve the image ratio following the different display cases. It will be the subject of another pull request 😉

PR #1218, opened 2023-10-12
Improve thumbnail in video progress bar

NicoPennec:
**Problem**
- On Edit page, there is no thumbnail on video progress bar.
- Too much styling calculation on mouseover progress bar.
- The tile image is not preloading of first init of component.

**Solution**
- Fix missing tile on video progress for Edit page.
- Refactor frame tile style in order to reduce compute.
- Improve tile preloading on init component.

### DIFF
--- a/src/components/pages/Edit.vue
+++ b/src/components/pages/Edit.vue
@@ -72,9 +72,10 @@
               :current-preview-index="currentPreviewIndex"
               :muted="isMuted"
               @max-duration-update="onMaxDurationUpdate"
+              @frame-update="onFrameUpdate"
               @metadata-loaded="onMetadataLoaded"
               @repeat="onVideoRepeated"
-              @frame-update="onFrameUpdate"
+              @video-loaded="onVideoLoaded"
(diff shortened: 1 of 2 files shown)


frankrousseau:
I'm a little bit concerned by this change. The natural dimensions led too many edge cases. That's why I decided to set the size on the backend size.

NicoPennec:
> I'm a little bit concerned by this change. The natural dimensions led too many edge cases. That's why I decided to set the size on the backend size.

After several tests, the dimension given by the video player still seems good to me. But I understand that you prefer to keep this logic on the back-end side.

The dimension given by the Kitsu API (Zou) can be wrong on some cases. So It will be necessary to modify the endpoint[1] to return the correct resolution based on the Display aspect ratio (DAR) for each preview video. (ping @EvanBldy)

[1] GET `/api/data/playlists/entities/{id}/preview-files`

frankrousseau:
It's better to ensure a proper computation on the backend side. The solution you propose requires to load the video before displaying the thumbnail.

NicoPennec:
I removed the commit that was causing debate.

PR #1222, opened 2023-10-18
Fix refreshing of production list

NicoPennec:
**Problem**
- On the Productions page, after reopening a production, the updated item is not visible in the open production list.

**Solution**
- Refactor the way to update the productions list to fix the reactivity issues.


frankrousseau:
@NicoPennec is this PR still relevant?

NicoPennec:
@frankrousseau Yes, but it's incomplete. I will finish it promptly!

PR #1239, opened 2023-11-13
Various fixes and improvements

frankrousseau:
**Problem**

The Miyu and Monelo studios reported a bunch of issues:

- The Pan/Zoom is active, while it shouldn't be when switching to fullscreen (it breaks the annotation canvas resizing).
- The priority is not displayed on done tasks, which makes things confusing when filtering and sorting.
- Departments, automations, and asset types are not archivable.
- Assignation filters do not apply to an asset type, which makes it hard to use.
- It's not possible to load more results on the search page.

**Solution**

- Disable pan / zoom for 2s when switching to full screen.
- Always display the priority in task cells.
- Add the capability to archive departments, automation, and asset types.
- Add a task type parameter to the assignation filter
- Add a load more button on the search page

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,3 @@ tests/test-bundle.js
 # Custom
 TODO
 QA
-
-# Cache
-.eslintcache
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ tests/test-bundle.js
 # IDE
 .idea/
 .vscode/
+.eslintcache
 
 # Custom
 TODO
--- a/src/store/api/departments.js
+++ b/src/store/api/departments.js
@@ -5,18 +5,26 @@ export default {
     client.get('/api/data/departments', callback)
   },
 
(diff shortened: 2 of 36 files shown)


.gitignore
```diff
@@ -49,6 +49,7 @@ tests/test-bundle.js
 # IDE
 .idea/
 .vscode/
+.eslintcache
```
NicoPennec, line 52: duplicate (already declared in the last line)

src/store/api/departments.js
```diff
@@ -5,18 +5,26 @@ export default {
...
-      `/api/data/departments/${editedDepartement.id}`,
-      editedDepartement
-    )
+  editDepartement(department) {
+    const data = {
+      name: department.name,
+      color: department.color,
+      archived: department.archived === 'true'
+    }
+    console.log(data)
```
NicoPennec, line 23: remove log

NicoPennec (approved):

PR #1244, opened 2023-11-16
[playlist] Add a full playlist timeline

frankrousseau:
**Problem**

* Playlists cannot be browsed as a single video file.
* Builds cannot be played

**Solution**

* Add a timeline dedicated to playlists
* See all shots in the timeline
* Shots are colored by status
* A build can be played while matching the corresponding shot in the timeline
* timeline preview available for the whole playlist


src/components/pages/playlists/RawVideoPlayer.vue
```diff
@@ -461,13 +469,16 @@ export default {
 
     runSetCurrentTime(currentTime) {
       const isChromium = !!window.chrome
-      const change = isChromium ? this.frameDuration : 0
+      const change = isChromium ? 0 : 0
```
NicoPennec, line 472: This condition always returns `0` ?

NicoPennec (approved):

src/components/pages/playlists/RawVideoPlayer.vue
frankrousseau: I removed it

PR #1248, opened 2023-11-23
[csv] add missing sanitize for some csv import + stop removing ';' an…

EvanBldy:
…d add always double quotes

**Problem**
- Some csv imports aren't sanitized before sending them to Zou.
- We remove delimiter ";" and don't double double quotes so it's not properly recognized by Zou (and it can leads to wrong behavior) 

**Solution**
- Sanitize every csv import before sending it to Zou. 
- Only double double quotes and always put the cell between double quote.

### DIFF
--- a/src/components/pages/Breakdown.vue
+++ b/src/components/pages/Breakdown.vue
@@ -925,8 +925,8 @@ export default {
     uploadImportFile(data) {
       const formData = new FormData()
       const filename = 'import.csv'
-      const csvcontent = csv.turnEntriesToCsvString(data)
-      const file = new File([csvcontent], filename, { type: 'text/csv' })
+      const csvContent = csv.turnEntriesToCsvString(data)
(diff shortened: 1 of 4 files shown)


NicoPennec (approved):

src/components/pages/Breakdown.vue
```diff
@@ -925,7 +925,8 @@ export default {
     uploadImportFile(data) {
       const formData = new FormData()
       const filename = 'import.csv'
-      const file = new File([data.join('\n')], filename, { type: 'text/csv' })
+      const csvcontent = csv.turnEntriesToCsvString(data)
+      const file = new File([csvcontent], filename, { type: 'text/csv' })
```
NicoPennec, line 929:
```suggestion
      const csvContent = csv.turnEntriesToCsvString(data)
      const file = new File([csvContent], filename, { type: 'text/csv' })
```

PR #1249, opened 2023-11-23
Improve CSV export

NicoPennec:
**Problem**
- The CSV export process uses a custom sanitization process.

**Solution**
- Use [Papa Parse](https://github.com/mholt/PapaParse/blob/master/README.md#parse-csv-with-javascript) to generate CSV files, to easily comply with the [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180).

### DIFF
--- a/src/lib/csv.js
+++ b/src/lib/csv.js
@@ -110,17 +110,14 @@ const csv = {
     return entries
   },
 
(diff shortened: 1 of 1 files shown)


EvanBldy:
Thanks @NicoPennec

PR #1255, opened 2023-11-29
[todos] Display the episodes where assets are casted

frankrousseau:
**Problem**

We can't see in the todo-list where the asset appears.

**Solution**

Add a column similar to the one used in the asset list showing the casted episodes.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -779,6 +779,7 @@ h2 {
   border: 1px solid $blue;
   border-radius: 5px;
   color: $blue;
+  cursor: pointer;
   font-size: 0.9em;
   padding: 3px 4px;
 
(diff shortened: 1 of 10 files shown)


src/components/lists/TodosList.vue
```diff
@@ -410,12 +421,44 @@ export default {
...
         Math.floor((this.timeEstimated ? this.timeEstimated : 0) / 60 / 8) <= 1
       )
+    },
+
+    isEpisodeVisible() {
+      return (
+        this.displayedTasks.filter(
+          task => task.source_id || task.episode_names?.length > 0
+        ).length > 0
+      )
```
NicoPennec, line 431: Prefer to use `some` instead of `filter`:

```suggestion
      return this.displayedTasks.some(
        task => task.source_id || task.episode_names?.length > 0
      )
```

PR #1261, opened 2023-12-11
Various fixes

frankrousseau:
**Problem**

* The asset list layout is clumsy.
* Annotations on pictures are broken when a video is displayed first.
* When getting into production, the asset page is always displayed by default. At some point in the production, it would be great that the shot page to become the default one.

**Solution**

* Put the loader and the no asset picture into the datable div wrapper
* Always use 0 as time for pictures annotations (saving / loading) and react to reset size event only if it comes from the right source in the player.
* Add the homepage field in the production parameters and use it to set the production links in the open production page.

### DIFF
--- a/src/locales/en.js
+++ b/src/locales/en.js
@@ -812,6 +812,7 @@ export default {
     open_productions: 'My Productions',
     picture: 'Change picture',
     title: 'Productions',
+
     creation: {
       add_assets: 'Add assets (optional)',
       add_asset_types: 'Add asset types',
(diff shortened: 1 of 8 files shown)


src/locales/en.js
```diff
@@ -904,6 +908,12 @@ export default {
       title: 'Brief'
     },
 
+    homepage: {
+      'assets': 'Assets',
+      'shots': 'Shots',
+      'sequences': 'sequences',
```
NicoPennec, line 914: typo:
```suggestion
      'sequences': 'Sequences',
```

PR #1280, opened 2024-01-12
Allow language variants depending on the production style

NicoPennec:
**Problem**
- Some terms can change depending on the production style (e.g. "shots" for animation, "level" for video game)

**Solution**
- Allow for language variations based on production style that extends default language.


frankrousseau:
Thank you!

PR #1286, opened 2024-01-17
Improve kanban board

NicoPennec:
**Problem**
- On kanban board, a user can update a task to any status without checking their rights.
- On kanban board, some task can require feedback before updating its status (e.g. WFA).
- On production settings, in the board section, selecting the roles to modify one by one is boring.
- With browsers based on Chromium, the cards are not draggable in the kanban board.

**Solution**
- Check that a user is authorized to update the status of a task.
- Add a confirmation modal if a task requires feedback.
- Add a button to select/unselect all roles on board settings.
- Fix event of card dragging for Chromium browser.

### DIFF
--- a/src/components/lists/KanbanBoard.vue
+++ b/src/components/lists/KanbanBoard.vue
@@ -18,10 +18,10 @@
       <li
         class="board-column"
         :key="column.id"
-        @dragenter="onCardDragEnter"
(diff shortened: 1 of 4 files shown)


PeteDraper:
additional - deleting a status to revert to a previous task doesn't update the board

PR #1298, opened 2024-01-22
Entity task list scroll issues

NicoPennec:
**Problem**
- The entity task list has scroll issues:
  - on the Edit, Episode, and Sequence pages, the list overflows if too many elements
  - on the Asset and Shot pages, the list is partially hidden if too many elements are displayed.

**Solution**
- Fix vertical overflow according to the page layout.


frankrousseau:
Thx!

PR #1305, opened 2024-01-23
[qa] Remove Canny

frankrousseau:
**Problem**

There is no guarantee that it is GDPR compliant and it adds a non-essential tracker to the application.

**Solution**

Remove the Canny widget and shows a link to our changelog instead.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -26,6 +26,5 @@ module.exports = {
     'vue/prop-name-casing': ['error', 'camelCase']
   },
   globals: {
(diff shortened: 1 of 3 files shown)


NicoPennec:
the references remains:
https://github.com/cgwire/kitsu/blob/969a5dee9df5718a6347344c6be9b864a3595a40/.eslintrc.cjs#L29
https://github.com/cgwire/kitsu/blob/969a5dee9df5718a6347344c6be9b864a3595a40/index.html#L20

PR #1316, opened 2024-01-25
Improve keyboard shortcuts on macOS

NicoPennec:
**Problem**
- on Apple macOS, some keyboard shortcuts don't work as expected. For example, in the preview player, you cannot copy/paste an annotation with the macOs conventional shortcut `Cmd + c` / `Cmd + v`, but it works with Windows/Linux conventional `Ctlr + c` / `Ctlr + v`.

**Solution**
- On key press, support the `Command` key (macOS) as a `Control` key (Windows/Linux)

### DIFF
--- a/src/components/mixins/player.js
+++ b/src/components/mixins/player.js
@@ -786,7 +786,7 @@ export const playerMixin = {
           event.stopPropagation()
           // ctrl + shift + left
           if (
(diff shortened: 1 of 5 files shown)


frankrousseau:
NB: There is probably a helper to add to the dom mixin.

PR #1333, opened 2024-02-09
Allow to set entity thumbnails automatically

frankrousseau:
**Problem**

https://cgwire.canny.io/asset-and-shot-tables/p/automatically-set-the-thumbnail-from-the-latest-preview-upload

**Solution**

Add a setting at the project level to allow to set automatically the latest upload as the entity thumbnail.

### DIFF
--- a/src/components/pages/production/ProductionParameters.vue
+++ b/src/components/pages/production/ProductionParameters.vue
@@ -52,7 +52,7 @@
           v-if="currentProduction && currentProduction.id"
         />
 
-        <text-field
+        <!--text-field
           ref="nbEpisodesField"
           type="number"
           :step="1"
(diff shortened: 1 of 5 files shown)


src/components/pages/production/ProductionParameters.vue
```diff
@@ -52,15 +52,15 @@
...
+        <!--text-field
           ref="nbEpisodesField"
           type="number"
           :step="1"
           :label="$t('productions.fields.nb_episodes')"
           @enter="runConfirmation"
           v-model="form.nb_episodes"
           v-if="currentProduction && currentProduction.id && isLocalTVShow"
-        />
+        /-->
```
NicoPennec, line 63: This field is no longer used?

frankrousseau: No

PR #1336, opened 2024-02-13
Fix kanban behavior if feedback is requested

NicoPennec:
**Problem**
- On the Kanban board, a task status with the flag "_is feedback request_" will request only one time to upload a preview file when updating the status.

**Solution**
-  Always ask for a preview file if the targeted status requires feedback.


frankrousseau:
Thx!

Issue #1331, opened 2024-02-07
Notification comment panel is not resizable


PeteDraper:
Kitsu version: 0.18.8

in the task view (eg shots) page you can expand the comments panel to make ti bigger or smaller, however in the Notifications page we can't amend the size. However, it does adopt the size that you last set the panel to, either in the task page or news page.

frankrousseau:
Thank you for the report Pete

NicoPennec:
Fixed on Kitsu >= 0.19.x

PR #1337, opened 2024-02-14
Task panel extendable by default

NicoPennec:
**Problem**
- On some pages, the task panel on the side is not extendable (related to https://github.com/cgwire/kitsu/issues/1331)

**Solution**
- Make the task panel extendable by default, and add an option to disable it.


frankrousseau:
Thx!

PR #1347, opened 2024-02-21
Various fixes

frankrousseau:
**Problem**

* Browsing from a TV show to a TV show empties the episode nav dropdown.
* In My Checks playlists the handles are  broken

**Solution**

* Properly set the production ID when production is changed and push the first episode to the browser route.
* Manage the case where handles are not set.

### DIFF
--- a/src/components/pages/playlists/PlaylistPlayer.vue
+++ b/src/components/pages/playlists/PlaylistPlayer.vue
@@ -1944,8 +1944,8 @@ export default {
       if (this.playlist.for_entity === 'shot') {
         entity = entity || this.currentEntity
         const shot = this.shotMap.get(entity.id)
-        this.handleIn = shot.data.handle_in || 0
-        this.handleOut = shot.data.handle_out || this.nbFrames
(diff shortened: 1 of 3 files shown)


src/components/pages/playlists/PlaylistPlayer.vue
```diff
@@ -1944,8 +1944,14 @@ export default {
...
-        this.handleIn = shot.data.handle_in || 0
-        this.handleOut = shot.data.handle_out || this.nbFrames
+        if (!shot) {
+          this.handleIn = 0
+          this.handleOut = this.nbFrames
+        } else {
+          if (!shot.data) shot.data = {}
+          this.handleIn = shot.data.handle_in || 0
+          this.handleOut = shot.data.handle_out || this.nbFrames
+        }
```
NicoPennec, line 1954: (proposal)
with optional chaining:

```js
          this.handleIn = shot?.data?.handle_in || 0
          this.handleOut = shot?.data?.handle_out || this.nbFrames
```

PR #1350, opened 2024-02-21
Various fixes

NicoPennec:
**Problem**
- On the Team schedule, a button is not translated.
- A TV Show production without episode has an invalid URL.

**Solution**
- Add the missing translation key.
- Set the episode to "all" by default to fix URL.


frankrousseau:
Thx!

PR #1368, opened 2024-03-05
Open tasks page

frankrousseau:
**Problem**

There is no way to browse tasks among all projects at the same time.

**Solution**

Adding a page listing all the tasks of all the open productions. It comes with some filters to quickly browse them.


frankrousseau:
I have to finish the server part prior to merging this

PR #1379, opened 2024-03-13
Fix episode initialization

NicoPennec:
**Problem**
-The episode and search can be reset when opening a new link to a shot.

**Solution**
- Fix episode initialization: search if an episode exists and keep the query search, or apply a default one and reset the search.


frankrousseau:
Thx!

PR #1386, opened 2024-03-19
[previews] Remove clumsy annotation reloading

frankrousseau:
**Problem**

In playlist annotations disappear when the UI is moved.

**Solution**

Remove unused annotation reloading.

### DIFF
--- a/src/components/mixins/player.js
+++ b/src/components/mixins/player.js
@@ -888,14 +888,10 @@ export const playerMixin = {
 
     reloadAnnotations(current = true) {
       if (!this.annotations) return
-      const annotations = this.annotations.map(a => ({ ...a }))
-      this.annotations = []
-      setTimeout(() => {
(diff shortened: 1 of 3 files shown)


src/components/mixins/player.js
```diff
@@ -888,22 +888,17 @@ export const playerMixin = {
...
       if (!this.annotations) return
-      const annotations = this.annotations.map(a => ({ ...a }))
-      this.annotations = []
-      setTimeout(() => {
-        this.annotations = annotations
-        if (current) {
-          this.reloadCurrentAnnotation()
-        }
-      }, 200)
+      this.annotations = this.annotations.map(a => ({ ...a }))
```
NicoPennec, line 891: This line seems useless, it's a shallow copy of himself.

frankrousseau: It's done to force Vue reactivity

Issue #1387, opened 2024-03-20
Unable to join translation


uuiid:
**Describe the bug**
I want to contribute to translation, but I am unable to join

**Screenshots**

![image](https://github.com/cgwire/kitsu/assets/46096697/54ece8a7-33d0-4b74-894d-734639dfe2fb)

frankrousseau:
Hello @uuiid, thank you for reporting. We'll check that soon.

frankrousseau:
It seems to be fixed.

PR #1395, opened 2024-03-28
Update Chinese translation

uuiid:
**Problem**

Outdated Chinese translation

**Solution**

Translate into Chinese


frankrousseau:
Thank you  but translations should be managed via POEditor. We'll see how to import it.

uuiid:
#1387 
Thank you, it's just that reason, so I opened a pull request

Issue #1398, opened 2024-03-29
Increase the number of productions visible in the pull-down menu to a larger amount, or give ability for the user to define this


PeteDraper:
Currently the pull-down menu in the Task page (etc) only gives a maximum of 5 visible entires - any more and you have to scroll down. Please increase this to a larger number or give the user the ability to amend the amount visible. 

Additionally, having the ability to jump to a project by hitting the character of the start of the project, eg T for Toy Story would jump to that entry. Would also be useful in the Productions page esp with a lot of active & inactive projects

PeteDraper:
The Productions list works great - seems to be set to 1/2 the height of the page. If possible, please apply the same logic to the Episodes pull-down menu

frankrousseau:
It will be fixed in the next release.

PR #1400, opened 2024-04-02
Improve the top bar menu

NicoPennec:
**Problem**
- The height of the production list in the top bar may be too small (related to #1398).
- As a Client user, the studio logo is not displayed in the top bar.

**Solution**
- Increase production list height based on screen size (max 50% of the viewport)
- Always display the studio logo in the top bar, but remain clickable if not a Client user.

### DIFF
--- a/src/components/tops/TopbarProductionList.vue
+++ b/src/components/tops/TopbarProductionList.vue
@@ -193,7 +193,7 @@ export default {
   border-bottom-left-radius: 5px;
   border-bottom-right-radius: 5px;
   margin-left: -6px;
-  max-height: 210px;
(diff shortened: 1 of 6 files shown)


src/components/tops/TopbarProductionList.vue
```diff
@@ -193,7 +193,7 @@ export default {
   border-bottom-left-radius: 5px;
   border-bottom-right-radius: 5px;
   margin-left: -6px;
-  max-height: 210px;
+  max-height: 50vh;
```
frankrousseau, line 196: This number should be a precise pixel number. Can you change it to 420px or similar?

PR #1406, opened 2024-04-04
Improve the team creation experience

frankrousseau:
**Problem**

It's painful to add people to a team.

**Solution**

* Add a button to import a team from another project
* Add a button to import an entire department
* Add a list to quickly add people from there


PeteDraper:
this is great! Lets do the same for linking in assets from other projects!!

PR #1437, opened 2024-05-13
[producions] Add a button to display production statistics

frankrousseau:
**Problem**

There is no way to see productions progress from the Productions page.

**Solution**

Add a button to load and display production statistics.

image


src/components/pages/production/ProductionStats.vue
```diff
@@ -0,0 +1,252 @@
...
+        taskTypeStatsMap[taskTypeStat.task_type_id].task_statuses.push(
+          taskTypeStat
+        )
+      })
+      return taskTypeStatsMap
+    },
+
+    statsByStatus() {
+      const statsByStatus = {}
+      this.stats.task_types.forEach(taskTypeStat => {
```
NicoPennec, line 172: `this.stats.task_types` can be undefined if no stats are returned from the API for an empty production.

PR #1474, opened 2024-07-02
[shots] Allow to set frames from previews

frankrousseau:
**Problem**

There is no way to compute frames from previews information.

**Solution**

Add a modal in the Shots page that allows to compute the number of frames from the previews of a given task type.

### DIFF
--- a/src/components/modals/SetFramesFromTaskTypePreviewsModal.vue
+++ b/src/components/modals/SetFramesFromTaskTypePreviewsModal.vue
@@ -1,0 +1,119 @@
+<template>
+  <div
+    :class="{
+      modal: true,
+      'is-active': active
+    }"
+  >
+    <div class="modal-background" @click="$emit('cancel')"></div>
+
+    <div class="modal-content">
+      <div class="box content">
+        <h1 class="title">
+          {{ $t('shots.get_frames_from_previews') }}
+        </h1>
+
+        <p class="description">
+          {{ $t('shots.get_frames_from_previews_description') }}
+        </p>
+
+        <combobox-task-type
+          :task-type-list="productionShotTaskTypes"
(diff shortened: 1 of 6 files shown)


src/components/modals/SetFramesFromTaskTypePreviewsModal.vue
```diff
@@ -0,0 +1,121 @@
...
+
+    isFormFilled() {
+      return this.taskTypeId !== null && this.taskTypeId !== ''
+    }
+  },
+
+  methods: {
+    ...mapActions([]),
+
+    addValue() {},
```
NicoPennec, line 96: empty method

src/components/modals/SetFramesFromTaskTypePreviewsModal.vue
```diff
@@ -0,0 +1,121 @@
...
+
+<script>
+import { mapGetters, mapActions } from 'vuex'
+import { modalMixin } from '@/components/modals/base_modal'
+
+import ComboboxTaskType from '@/components/widgets/ComboboxTaskType'
+import ModalFooter from '@/components/modals/ModalFooter'
+
+export default {
+  name: 'get-frames-from-previews-modal',
```
NicoPennec, line 48: wrong component name

```suggestion
  name: 'set-frames-from-task-type-previews-modal',
```

src/components/widgets/ComboboxTaskType.vue
```diff
@@ -80,6 +80,10 @@ export default {
       default: false,
       type: Boolean
     },
+    placeholder: {
+      default: '+ Task Type',
```
NicoPennec, line 84: Prefer to use an i18n key as default:

```js
      default: function () {
        return `+ ${this.$t('key')}`
      },
```

PR #1478, opened 2024-07-05
Various fixes

NicoPennec:
**Problem**
- On the New Production form:
  - the start date can be higher than the end date
  - the end date can be a weekend
  - an archive task type can be selected
- `slot` and `slot-scope` are deprecated on vue 2/3.
- ESLint configuration can make unintentional commits.

**Solution**
-  On the New Production form:
  - hide archived task types from selections.
  - fix selectable end dates.
- Use the v-slot shorthand notation.
- Update ESLint configuration:
  - disable auto-fix on commit.
  - update rules from `vue/essential` to `vue/recommended` (to be future-proof with vue 3)
  - fix ignore config.


PeteDraper:
The end date (and start date for that matter) should also include a weekend. Please don’t restrict to only weekdays

NicoPennec:
> The end date (and start date for that matter) should also include a weekend. Please don’t restrict to only weekdays

@PeteDraper We have considered your comment here: #1482

PR #1484, opened 2024-07-09
[breakdown] Fix header and colum styles

frankrousseau:
**Problem**

* Breakdown headers didn't display properly
* Breakdown column size didn't match header size

**Solution**

* Set a minimum height to the breakdown headers
* Use same default size for column and headers


EvanBldy:
Thanks :)

PR #1491, opened 2024-07-15
Various improvements

frankrousseau:
**Problem**

* Annotation group pasting is broken
* When a video with the wrong number of frames is uploaded, there is no warning.
* Following lucide-vue additions, frame number import icon and bot icon

**Solution**

* Ungroup pasted annotations
* Show a warning when the duration of uploaded videos doesn't match the number of frames information stored at the shot level
* Choose file digit and bot icons from lucide-vie to replace other icons

### DIFF
--- a/src/components/widgets/ButtonSimple.vue
+++ b/src/components/widgets/ButtonSimple.vue
@@ -45,7 +45,12 @@
     <codepen-icon class="icon" v-else-if="icon == 'codepen'" />
     <link-icon class="icon" v-else-if="icon == 'link'" />
     <clock-icon class="icon" v-else-if="icon == 'clock'" />
-    <file-digit stroke-width="1.2" size="20" class="icon" v-else-if="icon == 'file-digit'" />
+    <file-digit
+      stroke-width="1.2"
+      size="20"
+      class="icon"
+      v-else-if="icon == 'file-digit'"
+    />
     <kitsu-icon
       class="icon"
       :name="icon"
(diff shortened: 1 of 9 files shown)


src/components/widgets/ButtonSimple.vue
```diff
@@ -73,6 +79,7 @@ import {
   CornerLeftDownIcon,
   CornerRightDownIcon,
   DownloadIcon,
+  FileDigit,
```
NicoPennec, line 82: Use `FileDigitIcon` to be consistent with other icon imports
```suggestion
  FileDigitIcon,
```

src/components/widgets/ButtonSimple.vue
```diff
@@ -116,6 +123,7 @@ export default {
     DownloadIcon,
     EditIcon,
     Edit2Icon,
+    FileDigit,
```
NicoPennec, line 126:
```suggestion
    FileDigitIcon,
```

src/components/widgets/ButtonSimple.vue
```diff
@@ -45,6 +45,12 @@
     <codepen-icon class="icon" v-else-if="icon == 'codepen'" />
     <link-icon class="icon" v-else-if="icon == 'link'" />
     <clock-icon class="icon" v-else-if="icon == 'clock'" />
+    <file-digit
```
NicoPennec, line 48: Use `file-digit-icon` to be consistent with other icon imports.
It is helpful to quickly find all the `*-icons` components.

```suggestion
    <file-digit-icon
```

src/locales/en.js
```diff
@@ -1387,6 +1387,7 @@ export default {
     sequences: 'Sequences',
     tasks: 'Shot Tasks',
     title: 'Shots',
+    wrong_file_duration: 'One of the upldoaded video file duration doesn\'t match the expected duration of the current shot.',
```
NicoPennec, line 1390: Add variants in `en_nft.js` and `en_video-games.js`

> shot = NFT
> shot = map

EvanBldy:
Thanks

PR #1504, opened 2024-08-01
Minor fixes

frankrousseau:
**Problem**

* The concept side bar can't be displayed in the asset page
* Assets are not properly loaded when getting into the assets page from the production list

**Solution**

* Show the side bar when a concept is selected.
* Set a default value to the current section in the top bar

### DIFF
--- a/src/components/widgets/ConceptCard.vue
+++ b/src/components/widgets/ConceptCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="concept-item">
+  <div class="concept-item" @click="$emit('click')">
     <entity-preview
       :entity="concept"
       :empty-height="200"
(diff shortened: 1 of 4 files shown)


NicoPennec (changes requested):
The following commit "[[concepts] Display concept side bar in the asset page](https://github.com/cgwire/kitsu/pull/1504/commits/60248732011a81aea76b6cf8bb068f44fb6d9831)" is not consistent.
Changes related to the `@click` action from ConceptCard.vue are missing?

frankrousseau:
@NicoPennec I don't understand what you mean. Can you make it clearer?

src/components/widgets/ConceptCard.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="concept-item">
+  <div class="concept-item" @click="$emit('click')">
```
NicoPennec, line 2: This `$emit('click')` seems useless, as it was never used in the application.

frankrousseau: Ok thank you, a commit was missing.

EvanBldy:
Thanks !

PR #1519, opened 2024-08-29
Introduce SAML authentication

EvanBldy:
**Problem**
- It's not possible to do a SAML authentication in Kitsu.

**Solution**
- Add the possibility of doing SAML authentication.

### DIFF
--- a/src/components/pages/Login.vue
+++ b/src/components/pages/Login.vue
@@ -13,6 +13,19 @@
           <img src="../../assets/kitsu-text.svg" v-else />
         </div>
         <form v-if="!(isMissingOTP || isWrongOTP)">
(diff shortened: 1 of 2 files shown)


src/components/pages/Login.vue
```diff
@@ -13,6 +13,19 @@
...
         </div>
         <form v-if="!(isMissingOTP || isWrongOTP)">
+          <div class="field" v-if="mainConfig?.saml_enabled">
+            <p class="control">
+              <a
+                class="button is-fullwidth"
+                :class="{
+                  'is-loading': isLoginLoading
+                }"
+                @click="loginSAML"
```
NicoPennec, line 23: You can use a native link directly instead of a click event:

```suggestion
                href="/api/auth/saml/login"
```

PR #1525, opened 2024-09-05
Keep task id in url in entity task lists (+ fixes and updated japanese translation)

frankrousseau:
**Problem**

It's hard to navigate back from a task to an entity task list.

**Solution**

* Store the selected task ID in the URL query. 
* On the mount of the assets/shots/sequences/episodes/edits lists, select the given task ID. 
* Add a back button to the task page.

\+ Some dark theme fixes
\+ Update Japanese translation


src/lib/array.js
```diff
@@ -4,3 +4,5 @@ export const intersection = arrays => {
   rest = rest.map(x => new Set(x))
   return first.filter(x => rest.every(a => a.has(x)))
 }
+
+export const flatten = arr => arr.reduce((acc, val) => acc.concat(val), [])
```
NicoPennec, line 8: A `flat()` method already exists as a native feature in ES6.

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

PR #1529, opened 2024-09-12
[assets] Improve library ux

frankrousseau:
**Problem**

The current asset library experience is sometimes confusing.

**Solution**

Simplify asset library import: 
* select context first production and asset type to filter the available asset list.
* Always show the import column
* Use better color (orange looks like a warning) to show asset library elements

### DIFF
--- a/src/components/sides/ManageLibrary.vue
+++ b/src/components/sides/ManageLibrary.vue
@@ -7,7 +7,6 @@
       v-if="extendable"
     ></div>
     <div class="side manage-library">
-      <h2 class="mt0">{{ $t('library.manage') }}</h2>
       <div class="flexrowcolumn" v-if="selectedEntities.length">
         <delete-entities
           :error-text="$t('assets.multiple_delete_error')"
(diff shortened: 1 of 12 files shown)


src/components/sides/ManageLibrary.vue
```diff
@@ -233,10 +195,11 @@ export default {
     },
 
     productionUnsharedEntities() {
-      return this.unsharedAssets.filter(
-        entity => entity.project_id === this.productionId
-      )
+      return this.unsharedAssets
+        .filter(entity => entity.project_id === this.productionId)
+        .filter(entity => entity.entity_type_id === this.entityTypeId)
```
NicoPennec, line 200: Filters can be grouped together

```suggestion
      return this.unsharedAssets.filter(
        entity =>
          entity.project_id === this.productionId &&
          entity.entity_type_id === this.entityTypeId
      )
```

src/components/tops/ActionPanel.vue
```diff
@@ -314,7 +314,7 @@
             :title="$t('menu.clear_selection')"
```
NicoPennec, line 314: invalid i18n key
```suggestion
            :title="$t('main.clear_selection')"
```

src/components/tops/ActionPanel.vue
```diff
@@ -314,7 +314,7 @@
...
             @click="clearSelection"
           >
-            <x-icon :title="$t('menu.clear_selection')" size="16" />
+            <x-icon :title="$t('main.clear_selection')" :size="16" />
```
NicoPennec, line 317: `title` attribute already set on parent `<div>`

```suggestion
            <x-icon :size="16" />
```

PR #1535, opened 2024-09-18
Add setting to use hours instead of days for durations and estimations

MaximeGrosmaire:
**Problem**
We have a mix of artist that can work more or less daily on a project so it's important for us to be able to estimate in hours.

**Solution**
I've added an organization-wide setting to toggle duration and estimates in hours. 
This will require a zou commit as well to persist the setting.

I'm more of a TD than a proper dev so this PR might be a bit of a mess but I hope this is helpful in at least getting the ball rolling for that feature.

### DIFF
--- a/src/components/lists/AssetList.vue
+++ b/src/components/lists/AssetList.vue
@@ -474,9 +474,17 @@
         v-show="displayedAssetsTimeSpent > 0 || displayedAssetsEstimation > 0"
       >
         ({{ formatDuration(displayedAssetsTimeSpent) }}
-        {{ $tc('main.days_spent', displayedAssetsTimeSpent) }},
+        {{
+          isDurationInHours()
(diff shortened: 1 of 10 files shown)


frankrousseau:
Thank you very much @MaximeGrosmaire for this contribution! We will review and integrate this change soon.

Next time, please open an issue before submitting code to be sure we are ready to take the contribution.

frankrousseau:
We figured out that the server part is missing. It will require more time to integrate your changes.

MaximeGrosmaire:
I've submitted the zou change as well now for you guys to test now.

Production stats, asset list and task list should all show duration/estimations in hours once the setting is toggled.

With that setting enabled, we also don't get the start or due date from the estimate in days since that might end up setting unrealistic deadlines for anyone that is part-time / casual hours.

PR #1551, opened 2024-10-07
Various improvements

frankrousseau:
**Problem**

* There is no way to set the difficulty of a task
* Depending on project configuration, the publish tab should be displayed by default to artists
* We can't know who is the author of a shared filter.
* It's not possible to set the Kitsu as dark by default for every one.
* Notifications are missing filters and could be neater.

**Solution**

* Add a difficulty field to tasks.
* Add a setting to show the publish tab by default for the artists.
* Add a title to show the author of a shared filter.
* Add a setting to set the dark theme as default.
* Add read / unread and watching / is not watching filters to the notification page.
* In notifications, show only main information by default and show comments on the selection or in show comments mode.
* Allow to mark notifications as read/unread and do not mark them as read automatically.


src/components/widgets/AddComment.vue
```diff
@@ -441,26 +441,35 @@ export default {
...
+            e.stopPropagation()
+          })
+        }
+      })
     })
   },
 
   computed: {
     ...mapGetters([
+      'currentProduction',
```
NicoPennec, line 472: The `currentProduction` value may be incorrect on global pages (My Tasks, My Checks, News Feed, ...).

You can use `this.productionMap.get(eventData.project_id)` to retrieve the production of the current task.

frankrousseau: Thx. It's fixed.

PR #1554, opened 2024-10-17
Various improvements

frankrousseau:
**Problem**

* There is no way to limitate a shared filter to a department
* It's not possible to make Kitsu dark by default

**Solution**

* Allow to specify a department_id on shared filters
* Add a setting at the instance level to set it dark by default or not

### DIFF
--- a/src/components/modals/EditSearchFilterModal.vue
+++ b/src/components/modals/EditSearchFilterModal.vue
@@ -156,7 +156,6 @@ export default {
           ...this.form,
           is_shared: this.form.is_shared === 'true'
         }
-        console.log(data)
         this.$emit('confirm', data)
       }
     }
--- a/src/components/modals/EditSearchFilterModal.vue
+++ b/src/components/modals/EditSearchFilterModal.vue
@@ -31,8 +31,16 @@
           <boolean-field
             :label="$t('main.is_shared')"
             v-model="form.is_shared"
-            v-if="isCurrentUserManager && currentProduction"
             @click="form.search_filter_group_id = null"
+            v-if="isCurrentUserManager && currentProduction"
+          />
+
+          <combobox-department
+            class="mt2"
+            :label="$t('main.department')"
+            :top="true"
+            v-model="form.department_id"
+            v-if="form.is_shared === 'true'"
           />
 
           <combobox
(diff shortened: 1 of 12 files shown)


src/components/modals/EditSearchFilterModal.vue
```diff
@@ -146,6 +156,7 @@ export default {
           ...this.form,
           is_shared: this.form.is_shared === 'true'
         }
+        console.log(data)
```
NicoPennec, line 159: Log to remove:

```suggestion
```

src/components/widgets/ComboboxDepartment.vue
```diff
@@ -173,6 +174,28 @@ export default {
...
+          top: '-200px',
+          bottom: '-90px',
+          'border-top-left-radius': '10px',
+          'border-top-right-radius': '10px',
+          'border-bottom-left-radius': '0px',
+          'border-bottom-right-radius': '0px'
+        })
+      }
+
+      console.log(data, this.top)
```
NicoPennec, line 197: Log to remove:
```suggestion
```

src/components/widgets/ComboboxDepartment.vue
```diff
@@ -173,6 +174,28 @@ export default {
...
+      }
+      if (this.top) {
+        Object.assign(data, {
+          top: '-200px',
+          bottom: '-90px',
+          'border-top-left-radius': '10px',
+          'border-top-right-radius': '10px',
+          'border-bottom-left-radius': '0px',
+          'border-bottom-right-radius': '0px'
+        })
```
NicoPennec, line 194: The `top` value is set twice:
- line "183" : IF "top" THEN 30px ELSE 37px
- line "188" : IF "top" THEN -200px

The "30px" will never be applied.

```suggestion
      const data = {
        'max-height': this.maxHeightSelectInput + 'px',
        width: this.width + 'px',
        top: '37px',
        left: '0'
      }
      if (this.top) {
        Object.assign(data, {
          top: '-200px',
          bottom: '-90px',
          'border-top-left-radius': '10px',
          'border-top-right-radius': '10px',
          'border-bottom-left-radius': '0',
          'border-bottom-right-radius': '0'
        })
```

PR #1561, opened 2024-10-25
Improve Stats pages

NicoPennec:
**Problem**
In the Stats pages,
- The column order may be wrong on Kitsu or CSV export file.
- Export button will fail while loading data.
- The export of Retake stats may be wrong with empty cells.
- A long header name may overflow the column.

**Solution**
- Fix the column order in stats pages, based on production or studio settings.
- Disable the export button during stats loading.
- Fix the Retake stats export with data on matching columns in the CSV file.
- Improve style of table headers with long text (ellipsis + title).


frankrousseau:
Thank you!

Issue #1569, opened 2024-10-30
Allow editing estimates, durations and due dates from artist todo list


MaximeGrosmaire:
We like to run through each artist's todo list in our stand up meetings, would be great if we could update estimates, durations and due dates from there.

I'm happy to make a PR for this.

frankrousseau:
Hi @MaximeGrosmaire,

A PR for this would be welcome!

frankrousseau:
PR merged

PR #1583, opened 2024-11-13
Enhance TodosList to support task editing

MaximeGrosmaire:
**Problem**
We can't edit estimates and dates tasks in a user's todo list which means updating est and dates for all task assigned to a specific artist can require a lot of clicking around kitsu.

**Solution**
Make the estimates, start dates and due dates in the user's todo list editable. 

Issue: #1569

### DIFF
--- a/src/components/lists/TodosList.vue
+++ b/src/components/lists/TodosList.vue
@@ -41,21 +41,16 @@
             >
               {{ $t('assets.fields.description') }}
             </th>
-            <th
-              ref="th-estimation"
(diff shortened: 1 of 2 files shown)


NicoPennec:
Thanks for your contribution @MaximeGrosmaire 👍
We will integrate it soon, but we need to complete a technical migration upstream first.

NicoPennec:
Hi @MaximeGrosmaire
I integrated your proposal, after some corrections following the migration to Vue 3.
Thanks again

PR #1590, opened 2024-11-22
Improve copy comments from client

NicoPennec:
**Problem**
- On the Comments panel, the action to copy comments from a client is not fully complete, as the attachment files are not included.

**Solution**
- Improve the copy of comments from a client
  - Fix checklist unwanted reactivity
  - Duplicate attachment files


frankrousseau:
We do maintainance onlo on the `main` branch.

PR #1603, opened 2024-12-12
Various fixes

frankrousseau:
**Problem**

* Pinned revisions can be confused with being the last revision.
* Task columns with no department are hidden when a department filter is selected in entity pages.
* Shots loading doesn't occur when the episode is changed outside of the shot page.
* The set current frame as thumbnail option selects the previous frame instead of the current frame.
* Task type stats is not properly displayed in the production page.

**Solution**

* Change revision label to "Pinned Revision" for revisions in pinned comments
* Include task columns with no department in the selected column when a department is selected
* Check if the episode changed on Shots page mount.
* Add 1 frame to the selected frame to match ffmpeg way of counting frames
* Use the right number to compute the width of colored bars showing the amount of tasks for each status

### DIFF
--- a/src/components/mixins/entity_list.js
+++ b/src/components/mixins/entity_list.js
@@ -428,6 +428,7 @@ export const entityListMixin = {
     validationColumnsIsInDepartmentFilter(columnId) {
       return (
         this.departmentFilter.length === 0 ||
+        this.taskTypeMap.get(columnId).department_id === null ||
         this.departmentFilter.includes(
           this.taskTypeMap.get(columnId).department_id
         )
--- a/src/components/mixins/entity_list.js
+++ b/src/components/mixins/entity_list.js
@@ -428,7 +428,7 @@ export const entityListMixin = {
     validationColumnsIsInDepartmentFilter(columnId) {
       return (
         this.departmentFilter.length === 0 ||
-        this.taskTypeMap.get(columnId).department_id === null ||
+        this.taskTypeMap.get(columnId)?.department_id === null ||
         this.departmentFilter.includes(
           this.taskTypeMap.get(columnId).department_id
         )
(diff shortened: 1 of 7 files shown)


src/components/sides/TaskInfo.vue
```diff
@@ -1315,7 +1315,7 @@ export default {
         this.loading.setFrameThumbnail = true
         let frame = null
         if (isUseCurrentFrame) {
-          frame = this.currentFrameRaw
+          frame = this.currentFrameRaw + 1
```
NicoPennec, line 1318: Please add a comment on this line to explain why "+ 1".
It will be easier to read and understand the code later (git blame is not viable over time)

src/components/widgets/Comment.vue
```diff
@@ -278,7 +278,13 @@
...
           :to="previewRoute"
         >
-          {{ $t('comments.revision') }} {{ comment.previews[0].revision }}
+          <template v-if="comment.pinned">
+            {{ $t('comments.pinned_revision') }}
+            {{ comment.previews[0].revision }}
+          </template>
+          <template v-else>
+            {{ $t('comments.revision') }} {{ comment.previews[0].revision }}
+          </template>
```
NicoPennec, line 287: alternative:
```suggestion
          {{
            comment.pinned
              ? $t('comments.pinned_revision')
              : $t('comments.revision')
          }}
          {{ comment.previews[0].revision }}
```

src/components/mixins/entity_list.js
```diff
@@ -428,6 +428,7 @@ export const entityListMixin = {
     validationColumnsIsInDepartmentFilter(columnId) {
       return (
         this.departmentFilter.length === 0 ||
+        this.taskTypeMap.get(columnId).department_id === null ||
```
NicoPennec, line 431: `this.taskTypeMap.get(columnId)` will never be empty ?

else add an optional chaining:
```suggestion
        this.taskTypeMap.get(columnId)?.department_id === null ||
```

PR #1605, opened 2024-12-13
[project] always use 25 fps as default value

EvanBldy:
**Problem**
- We don't always use 25 fps as the default value.

**Solution**
- Use 25 fps as default value everywhere.


frankrousseau:
Why do you use 25 over 24?

NicoPennec (approved):

Issue #1610, opened 2024-12-14
Fatal - Changing from Episodes Stats to Sequence Stats Causes browser page to completely lock up


PeteDraper:
**Context**

Studio name: Ryff
Kitsu version: 0.20.1
Production type: TV series

Chrome Browser

Changing from Episodes stats to Sequence stats causes the entire browser page to lock up - can't even for refresh with Command+R (OSX).

PeteDraper:
Actually this occurs when changing from other stats / pages in the project as well, eg news feed.

NicoPennec:
Thank you @PeteDraper 🙏 
Fixed on Kitsu >= v0.20.2

Issue #1609, opened 2024-12-14
Changing projects clears out the episodes list. Selecting an episode says no shots present. Page refresh needed


PeteDraper:
**Context**

Studio name: Ryff
Kitsu version: 0.20.1
Production type: 

While inside a project, selecting another project results in the Episodes list defaulting to a blank entry with the green animated logo in the page just spinning endlessly. 

Pulling down the episodes menu gives the correct list of episodes for that project, however selecting one results in a page saying "no shots present - how about creating some?". A page refresh is needed to display the shots.

NicoPennec:
Thank you @PeteDraper 🙏 
Fixed on Kitsu >= v0.20.2

PR #1614, opened 2024-12-16
Fix navigation

NicoPennec:
**Problem**
- #1610
- #1609
- Wrong episode ID on the production links of the top bar.

**Solution**
- Avoid an infinite loop on navigation by checking if the episode ID has really changed.
- Always clear the current episode to ensure the correct watching event on the ID change.
- Force episode "all" on the path when switching projects.


frankrousseau:
Thank you for the fixes. I have two remarks:

- Having the 'all' episode by default doesn't make sense for shots/sequences/episode/edits where we need the first episode selected by default.
- Clearing episodes is important or some UIs won't refresh

NicoPennec:
> Thank you for the fixes. I have a single remark:
> 
> * Having the 'all' episode by default doesn't make sense for shots/sequences/episode/edits where we need the first episode selected by default.

Got it. I removed that change.

PR #1631, opened 2024-12-20
Various improvements

frankrousseau:
**Problem**

* Scrolling wide entity lists via scrollbars is a little bit painful
* Pictures are always reloaded in playlists
* Some locales appear on several lines in the sidebar

**Solution**

* Solution test: Allow to grab the shot list to scroll it
* In playlists prepare all pictures and switch between them
* Change font size in the sidebar for Japanese and French locales

### DIFF
--- a/src/components/lists/ShotList.vue
+++ b/src/components/lists/ShotList.vue
@@ -249,6 +249,8 @@
           <tbody
             class="datatable-body"
             :key="getGroupKey(group, k, 'sequence_id')"
+            @mousedown="startBrowsing"
+            @touchstart="startBrowsing"
             v-for="(group, k) in displayedShots"
           >
             <tr class="datatable-type-header">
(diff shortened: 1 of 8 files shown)


NicoPennec:
Another approach to the sidebar is to set its width to be dynamic, as below:

```css
aside {
  /* width: 230px; */
  min-width: 230px;
  max-width: 350px;
}

aside.hidden-bar {
  left: -350px;
}
```

src/components/lists/ShotList.vue
```diff
@@ -805,10 +807,30 @@ export default {
...
+        ['touchend', this.stopBrowsing],
+        ['touchcancel', this.stopBrowsing],
+        ['keyup', this.stopBrowsing]
+      ]
     }
   },
 
+  mounted() {
+    this.stickedColumns =
+      JSON.parse(localStorage.getItem(this.localStorageStickKey)) || {}
```
NicoPennec, line 825: Maybe use the `getObjectPreference` helper from `preference.js`

frankrousseau:
> Another approach to the sidebar is to set its width to be dynamic, as below:
> 
> ```css
> aside {
>   /* width: 230px; */
>   min-width: 230px;
>   max-width: 350px;
> }
> 
> aside.hidden-bar {
>   left: -350px;
> }
> ```

I don't want to change the studio logo size from an instance to another and the current width is matching the studio logo size (+ margins).

PR #1636, opened 2024-12-29
updates frame number on save to match player.js frame

tetsuoanimation:
**Problem**
annotations saved by preview player was off by one frame compared to generalized player.js implementation
**Solution**
Add 1 to the annotation current frame on annotation save


tetsuoanimation:
This was meant to be merged into our own main first - will send a full pr later

Issue #1595, opened 2024-12-02
Annotation Data is inconsistent between Preview Player and Playlist Player


tetsuoanimation:
**Context**

Studio name: Tetsuo Animation
Kitsu version: Kitsu 0.19.74
Production type: Short

**Describe the bug**

When adding a drawing on a film, the frame data is inconsitent depending on the frame being drawn in PreviewPlayer mode or Playlist Mode. It shows up at the 'correct' time and @frame links seem to work in both players.

Here's an example:
The first two datas have been drawn in PreviewPlayer mode - the frame numbers are saved as int and are off by one frame. ( First drawing is on Frame 55 on the framecounter ( second screenshot )

The second two datas have been drawn in PlaylistMode - the frame numbers are saved as string and are matching the clock ( third screenshot ) 

The clocks are the same in PreviewPlayer and PlaylistMode and I _think_ the times might be correct as the drawing is shown on the right frame. 

I also found that the PreviewPlayer shows a different image from the video depending on how we seek to it - stepping frame by frame is one frame of from seeking to a specific frame, no matter if we do it with the timeline or through code.
I'll investigate this a bit more and open a second issue for it.

**Screenshots**
![grafik](https://github.com/user-attachments/assets/2910f4fb-31c7-4375-9987-1ac9bbcdc271)
![grafik](https://github.com/user-attachments/assets/38e02be3-9a5d-4aae-ab71-28a4d6818435)
![grafik](https://github.com/user-attachments/assets/3858d3d3-44c5-4625-ac98-20c4b68237cb)

tetsuoanimation:
Checked if raw time saved by player.js and PreviewPlayer.vue is the same ✅ 
Checked which function is the library function ( player.js ) ✅ 
Tested / checked data integrity ✅ 

I made a very simple PR for this one - https://github.com/cgwire/kitsu/pull/1637

tetsuoanimation:
Patched in https://github.com/cgwire/kitsu/pull/1638

PR #1637, opened 2024-12-29
Fix for #1595  / Annotation Data is inconsistent between Preview Player and Playlist Player

tetsuoanimation:
**Problem**
Annotation frame number was off by one frame when saving from PreviewPlayer.vue compared to player.js

**Solution**
Add 1 to frame number when saving from PreviewPlayer.vue after checking and confirming that both ways save the same 'raw' time.


tetsuoanimation:
Fixed the issue only on the front end but not on DB

tetsuoanimation:
Now - new PR in https://github.com/cgwire/kitsu/pull/1638

PR #1641, opened 2024-12-31
Drawing Improvements | PSBrush Library for pressure sensitivity / Shortcuts to jump between drawings

tetsuoanimation:
**Problem**
There is no pressure sensitivity when drawing with a pen
There is no way to jump between drawings

**Solution**
Implement PSBrush library
Implement , and . to jump between drawings

**Note**
This might need changes from my end or might not be what you need. Feel free to reject.
After updating to the latest main from cgwire/kitsu, the playlists break, but it seems like that might come from a new feature / other problem in PictureViewer ( I can't verify that currently )


frankrousseau:
Thank you! That is going to be well appreciated. 

Just a question: why did you use this library instead of the one that allows to chose different brushes?

tetsuoanimation:
Sure! I checked the other libraries and I couldn't really find many others that support pressuresensitivity. I compared a few and found this to be the 'simplest':This library was:- simple ( easy to read / understand )- fewer features ( easier to maintain )- doesn't need extra GUI work- feels the same when used with a mouse- supports pressure sensitivity in a way that looked the easiest to implement into the current system without breaking existing features.- didn't need changes to how zou stores the serialised drawing data.- doesn't need opacity etc.Drawbacks:- github of the library dormant for a while now- only a single way of drawing / single brush- needed monkeypatching to work with the current cgwire/fabricWhat do you think? :)Von: Frank Rousseau ***@***.***>Gesendet: Dienstag, Dezember 31, 2024 1:52:03 PMAn: cgwire/kitsu ***@***.***>Cc: Tetsuo Animation ***@***.***>; Author ***@***.***>Betreff: Re: [cgwire/kitsu] Drawing Improvements | PSBrush Library for pressure sensitivity / Shortcuts to jump between drawings (PR #1641)
Thank you! That is going to be well appreciated.
Just a question: why did you use this library instead of the one that allows to chose different brushes?

frankrousseau:
Do you think that you can make your git history cleaner? It's a little bit hard to select the right commit to merge.

src/components/previews/PreviewPlayer.vue
```diff
@@ -1815,15 +1809,15 @@ export default {
         } else if (event.keyCode === 68) {
           // d
           this.container.focus()
-          this.pauseEvent(event)
+          this.pauseEvent(event)-1 //compensate for offsert frame number
```
frankrousseau, line 1812: This line doesn't make sense.

src/components/previews/PreviewPlayer.vue
```diff
@@ -1815,15 +1809,15 @@ export default {
...
           this.onPencilAnnotateClicked()
         } else if ((event.ctrlKey || event.metaKey) && event.keyCode === 90) {
           // ctrl + z
           this.undoLastAction()
         } else if (event.altKey && event.keyCode === 82) {
           // alt + r
           this.redoLastAction()
-        } else if (event.altKey && event.keyCode === 74) {
+        } else if (event.altKey && event.keyCode === 74) {-1 // compensate for offset frame number
```
frankrousseau, line 1820: This line doesn't make sense.

tetsuoanimation:
It seems like merging main went wrong somewhere - I'll get this fixed when I'm back on Monday.

src/components/previews/PreviewPlayer.vue
tetsuoanimation: I'll update with a proper rebase so the history is clean and send a new PR

src/components/previews/PreviewPlayer.vue
tetsuoanimation: I'll update with a proper rebase so the history is clean and send a new PR

Issue #1643, opened 2025-01-02
Asset Path UI Path Overlap


darkprince334455:
Context

Studio name: Monsta SDN BHD
Kitsu version: 0.20.11
Production type: TV Show

Describe the bug

Asset path text overlaps with other columns. It only happens when viewing the task as Artist and Supervisor.

Screenshots
![image](https://github.com/user-attachments/assets/5175bd6f-0def-4762-a488-5243f2be27b3)

Desktop (please complete the following information):

OS: [Windows 10 pro]
Browser [Chrome]
Version [131.0.6778.205 (Official Build) (64-bit)]

NicoPennec:
Issue fixed on Kitsu >= 0.20.13

Thank @darkprince334455

Issue #1629, opened 2024-12-19
Mark As Read button & individual notification status is broken


PeteDraper:
**Context**

Studio name: Ryff
Kitsu version: 0.20.8
In Notifications, the Mark As Read button has no effect, also the individual "Read" button of a Notification doesn't work - it checks the item as Read, but on page refresh that Read is back to grey.

NicoPennec:
Issue fixed on Kitsu >= 0.20.9

Thank you for the report 🙏

PeteDraper:
this still appears to be broken in 0.20.12

NicoPennec:
@PeteDraper the button works now, but it's still an UI issue I will fixed it. Thanks

NicoPennec:
The last UX issue is fixed on Kitsu >= 0.20.13

Issue #1626, opened 2024-12-18
Start and Due Date calendars should stick out, instead of being wrapped in a task column


makermatic:
**Context**

Studio name: PP
Kitsu version: 0.20.5
Production type: Feature Film

**Describe the bug**

When you select either date column in My Tasks, the calendar view looks like its wrapped into the task column. Which makes it tougher to pick a date. Instead it should stick out of the column like a pop up

**Screenshots**

![Recording 2024-12-18 at 13 51 09](https://github.com/user-attachments/assets/858266e1-68e7-4a8c-99f2-e64f12971943)

**Desktop (please complete the following information):**

 - OS: macOS
 - Browser: Edge
 - Version Sonoma 14.6

NicoPennec:
Issue fixed on Kitsu >= 0.20.13

Thanks @makermatic for the bug report.

makermatic:
Thanks Nico!!

PR #1644, opened 2025-01-02
Various fixes

NicoPennec:
**Problem**
- https://github.com/cgwire/kitsu/issues/1629
- https://github.com/cgwire/kitsu/issues/1643
- https://github.com/cgwire/kitsu/issues/1626

**Solution**
- Fix a wrong state on the "read" notification button (issue due to vue3 migration).
- Improve the display of metadata in read-only mode, in order to avoid UX overflow.
- Fix the visibility of the date picker if hidden overflow. Thanks to the "teleport" option from the date picker lib.

### DIFF
--- a/src/components/cells/MetadataInput.vue
+++ b/src/components/cells/MetadataInput.vue
@@ -2,6 +2,7 @@
   <!-- text input -->
   <input
     class="input-editor"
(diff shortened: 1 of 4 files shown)


frankrousseau:
Thx!

Issue #1648, opened 2025-01-06
Playlist Forever Loading


darkprince334455:
Context

Studio name: Monsta SDN BHD
Kitsu version: 0.20.11
Production type: TV Show

Describe the bug

Selecting a shot playlist will show a loading icon forever, even though the first shot is auto-selected for view. The workaround is to choose a shot for it to appear correctly. This will affect live room due to all participants needing to choose a shot to get it working

Screenshots
![image](https://github.com/user-attachments/assets/4fc4e1e5-4bb0-4852-9f89-faa658f7a380)

Desktop (please complete the following information):

OS: [Windows 10 pro]
Browser [Chrome]
Version [131.0.6778.205 (Official Build) (64-bit)]

NicoPennec:
Thank you @darkprince334455 
We have made some improvements in the latest release of the Playlist page.
Could you please retry again with Kitsu >= 0.20.14 ?

PR #1649, opened 2025-01-06
Cleaned up history for Drawing Improvements | PSBrush Library for pressure sensitivity / Shortcuts to jump between drawings

tetsuoanimation:
**Problem**
PR #1641 was unclear and had a messy history with multiple merges of the main branch.
PR #1641 had a broken merge that didn't make sense

Original PR below

**Solution**
- Rebased to latest upstream so history is easier to read.
- Fixed merging problems 1 by 1

Problem
- There is no pressure sensitivity when drawing with a pen
- There is no way to jump between drawings

Solution
- Implement PSBrush library
- Implement , and . to jump between drawings

Note
This might need changes from my end or might not be what you need. Feel free to reject.
After updating to the latest main from cgwire/kitsu, the playlists break, but it seems like that might come from a new feature / other problem in PictureViewer ( I can't verify that currently ) - this seems to be reported in #1648 by another user.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@animxyz/vue3": "0.6.7",
+        "@arch-inc/fabricjs-psbrush": "^0.0.20",
         "@fullcalendar/core": "6.1.15",
         "@fullcalendar/daygrid": "6.1.15",
         "@fullcalendar/multimonth": "6.1.15",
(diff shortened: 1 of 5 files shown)


frankrousseau:
I think you need to use `git pull --rebase origin main`. 

If you tell me which commit I should use, I can try to apply them directly on the main branch myself

tetsuoanimation:
If you get latest from the tetsuoanimation:tetsuo-feature-ps-brush [1633997](https://github.com/cgwire/kitsu/pull/1649/commits/16339979c9f0cbd82f840e815994975982d38f1a), that's the one.

tetsuoanimation:
An open-source experienced friend told me that it is better for projects if contributions are squashed by the contributors so maintainers have to deal with only a single change per feature. 
So.. let me squash rebase this to make it as easy as possible for you. I'll let you know when it's ready - so you don't have to deal with my back and forth and and comments / console logs - only commits as I was figuring out how the system works 
( I promise future contribs will be cleaner ;) )

PR #1655, opened 2025-01-09
Drawing Improvements | PSBrush Library for pressure sensitivity / Shortcuts to jump between drawings | Squash into Single Commit rebased on latest upstream

tetsuoanimation:
**Problem:**
- PR [1649](https://github.com/cgwire/kitsu/pull/1641) had a history / multiple commits

**Solution:**
- Squash merge into recreated branch so there's only a single commit for this feature.

_Original PR below_

**Problem**
- There is no pressure sensitivity when drawing with a pen
- There is no way to jump between drawings

**Solution**
- Implement PSBrush library
- Implement , and . to jump between drawings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@animxyz/vue3": "0.6.7",
+        "@arch-inc/fabricjs-psbrush": "^0.0.20",
         "@fullcalendar/core": "6.1.15",
         "@fullcalendar/daygrid": "6.1.15",
         "@fullcalendar/multimonth": "6.1.15",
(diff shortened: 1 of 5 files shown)


frankrousseau:
Thank you but I have one additional question. Why the `annotation.js` file is entirely rewritten. It's hard to review changes.

tetsuoanimation:
> Thank you but I have one additional question. Why the `annotation.js` file is entirely rewritten. It's hard to review changes.

I think that's trailing whitespace that my compare ignored - let me see what I can do.

frankrousseau:
It's the same for the PreviewPlayer file.

tetsuoanimation:
Yeah - ignoring whitespace when comparing works - adding ?w=1 on the link on github does that: https://github.com/cgwire/kitsu/pull/1655/files?w=1

frankrousseau:
Ok I will work on this on my code directly if you don't mind to avoid the unreadable diff in the git history.

tetsuoanimation:
> Ok I will work on this on my code directly if you don't mind to avoid the unreadable diff in the git history.

Oh totally fine :)

frankrousseau:
I'm closing this PR in favor of mine which is cleaner from a diff perspective. https://github.com/cgwire/kitsu/pull/1657 

Thank you a lot @tetsuoanimation for this great addition!

Issue #1654, opened 2025-01-08
Assets in Asset Library don't Wrap Text


makermatic:
**Context**

Studio name: Pudgy
Kitsu version: 0.20.15
Production type: Feature Film

**Describe the bug**

When scrolling through the asset library, I noticed some longer names overlap the text box instead of wrapping into a new line. The first two screenshots show the error and the third one shows a long name that gets close to touching the edge of the box.

**Screenshots**
<img width="210" alt="Screenshot 2025-01-08 at 9 37 40 AM" src="https://github.com/user-attachments/assets/bb034ea5-ef22-4736-b167-81d761b9452b" />
<img width="177" alt="image" src="https://github.com/user-attachments/assets/b0769c15-2d25-4c16-9f17-cfd46129854d" />
<img width="196" alt="image" src="https://github.com/user-attachments/assets/66d230ca-0b68-4e15-a12e-3b9c83f4c3cd" />

**Desktop (please complete the following information):**

 - OS: MacOS
 - Browser: Edge

NicoPennec:
Issue fixed in Kitsu >= 0.20.16

Thank you for your contribution @makermatic 🙏

PR #1656, opened 2025-01-09
Various fixes

NicoPennec:
**Problem**
- On the newsfeed page, some errors were raised due to read-only mode.
- Sometimes production settings are not applied well on the task panel.
- Long asset names may overflow into the UI (eg. https://github.com/cgwire/kitsu/issues/1654)
- Real-time updates on the departements settings page do not work on create and delete actions.
- On the task type page, switching to view as a contact sheet failed the first time.
- On the preview player, changing the current animation for the 3D model does not reset the playing status, which may be paused instead of playing.
- The playlist progress bar does not work with a touch event.
- Some values ​​can be undefined in specific cases and generate side effects or silent errors.
- Some outdated npm dependencies.

**Solution**
- Fix errors in read-only mode on the preview player in the newsfeed.
- Fix the current production on the following pages: my checks, my tasks, studio newsfeed, and notifications.
- Wrap long name in assets library and task type pages.
- Fix real-time updates for departments. Add missing getter.
- Fix initialization of the contact sheet view in the task type page.
- Fix preview playing status on animation change for 3D model.
- Fix the touch event on playlist progress bar. The method called was incorrect.
- Fix edge cases of undefined value.
- Bump npm dependencies (except Marked and JSDom for technical reasons).


frankrousseau:
Thank you!

PR #1657, opened 2025-01-12
Add pen pressure to annotations + various fixes

frankrousseau:
**Problem**

* The drawing tool doesn't support pen pressure
* Sequence stats loading doesn't occur when shots are already loaded

**Solution**

* Use PSBrush library to handle pen pressure and fake mouse pressure based on distance (or time) when using a mouse or a touch device. Solution implemented by @tetsuoanimation 
* Reload sequences if some data are missing


src/components/mixins/player.js
```diff
@@ -628,6 +628,48 @@ export const playerMixin = {
...
+          this.setFullPlayerTime(nextDrawingTime)
+        } else {
+          this.rawPlayer.setCurrentTimeRaw(annotation_time / this.fps)
+          this.onProgressChanged(annotation_time, true)
+        }
+        if (this.isComparing) {
+          this.syncComparisonViewer()
+        }
+      } catch {
+        console.log('should never happen')
```
NicoPennec, line 647: Improve error log to aid debugging:
```suggestion
      } catch (err) {
        console.error('should never happen', err)
```

src/components/mixins/player.js
```diff
@@ -628,6 +628,48 @@ export const playerMixin = {
...
```
NicoPennec, line 668: Improve error log to aid debugging:
```suggestion
      } catch (err) {
        console.error('should never happen', err)
```

src/components/mixins/player.js
```diff
@@ -974,6 +1038,41 @@ export const playerMixin = {
       if (annotation) this.loadAnnotation(annotation)
     },
 
+    getSortedAnnotations() {
+      const annotations = this.annotations
+      annotations.sort((a, b) =>
+        a.time > b.time ? 1 : b.time > a.time ? -1 : 0
+      )
```
NicoPennec, line 1045:
```suggestion
      annotations.sort((a, b) => a.time - b.time)
```

src/components/previews/PreviewPlayer.vue
```diff
@@ -1536,6 +1566,41 @@ export default {
       }
     },
 
+    getSortedAnnotations() {
+      const annotations = this.annotations
+      annotations.sort((a, b) =>
+        a.time > b.time ? 1 : b.time > a.time ? -1 : 0
+      )
```
NicoPennec, line 1573:
```suggestion
      annotations.sort((a, b) => a.time - b.time)
```

src/components/mixins/annotation.js
```diff
@@ -1038,12 +1144,124 @@ export const annotationMixin = {
...
+          : this.mouseDrawingMaxPressure
+      }
+    },
+
+    /*
+     * Get the distance between two fabric points.
+     */
+    getPointDistance(p1, p2) {
+      return Math.sqrt(Math.abs(p1.x - p2.x) + Math.abs(p1.y - p2.y))
+    },
```
NicoPennec, line 1166: never used on code ?

src/components/mixins/annotation.js
```diff
@@ -1038,12 +1144,124 @@ export const annotationMixin = {
...
+    /*
+     * Get the distance between two fabric points relative to the canvas size.
+     */
+    getCanvasRelativePointDistance(p1, p2, canvas) {
+      const dimensions = new fabric.Point(canvas.getWidth(), canvas.getHeight())
+      const p1_rel = p1.divide(dimensions)
+      const p2_rel = p2.divide(dimensions)
+      return Math.sqrt(
+        Math.abs(p1_rel.x - p2_rel.x) + Math.abs(p1_rel.y - p2_rel.y)
+      )
```
NicoPennec, line 1177: @tetsuoanimation I'm not sure to understand this one.
It doesn't look like a classic distance calculation (Euclidean or Manhattan distance formula).

frankrousseau: It's a Pythagore formula. With two coordinates and their projections, you can make a triangle rectangle.

frankrousseau: But you're right, it seems that squares are missing.

frankrousseau: I will let it as it is because it works better than real distance and add some comments.

src/components/mixins/player.js
```diff
@@ -643,8 +643,8 @@ export const playerMixin = {
         if (this.isComparing) {
           this.syncComparisonViewer()
         }
-      } catch {
-        console.log('should never happen')
+      } catch (err) {
+        console.err('wrong call from unexpected player', err)
```
NicoPennec, line 647:
```suggestion
        console.error('wrong call from unexpected player', err)
```

src/components/mixins/player.js
```diff
@@ -664,8 +664,8 @@ export const playerMixin = {
...
```
NicoPennec, line 668:
```suggestion
        console.error('wrong call from unexpected player', err)
```

src/components/mixins/annotation.js
NicoPennec: ok, notice that current usage of `pow(..., 1)` is useless.

Else, for the real distance you can use the [`Math.hypot()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot) operator:
`return Math.hypot(p1_rel.x - p2_rel.x, p1_rel.y - p2_rel.y)`

and remove this weird line:
`delta_dist *= 50 // magic number to scale to nicer values`

tetsuoanimation: > But you're right, it seems that squares are missing.

I think in this case, I always ended up with squaring a 1 which does nothing - but I might be wrong?

I found a note that `Math.hypot` is very slow - since this is basically running every frame I preferred the faster `sqrt` way. It's totally ok if you'd prefer it.

The `delta_dist` scaling is weird indeed and I don't like using a magic number there. It just makes the resulting line and pressure values look prettier on screen and I couldn't find another way to make it look as nice.

Totally happy to adjust.

PR #1665, opened 2025-01-20
[events] stop sending event preview-room:leave with temp as room_id

EvanBldy:
**Problem**
- When entering the playlist page it sends an event "preview-room:leave" with "temp" as room_id, causing an error in zou-events.

**Solution**
- Stop sending that event when no playlist was previously selected.


NicoPennec (approved):

EvanBldy:
> Why change the "early return" coding style to a "conditional block" style? The original style makes the code easier to read.

It's done :)

PR #1672, opened 2025-01-21
Various fixes

frankrousseau:
**Problem**

* Annotation attachment is no longer working when there are annotations with pen pressure.
* The attached annotations are not properly sorted.
* Task selection is lost on reload in entities page
* Search reload is broken when a task is unselected leading to not allowing to clear the search right away

**Solution**

* Add missing function to pressure path prototype
* Fix sorting function before extracting annotations
* Retrieve entityMap at the right time to allow task selection
* Apply search from route only if no task is selected

### DIFF
--- a/src/components/widgets/SearchField.vue
+++ b/src/components/widgets/SearchField.vue
@@ -53,11 +53,6 @@ export default {
       type: Boolean,
       default: false
     },
-    // FIXME: property no longer used in component
-    active: {
-      type: Boolean,
-      default: true
-    },
     focusOptions: {
       type: Object
     }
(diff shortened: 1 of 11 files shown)


src/components/widgets/SearchField.vue
```diff
@@ -53,11 +53,6 @@ export default {
       type: Boolean,
       default: false
     },
-    // FIXME: property no longer used in component
-    active: {
-      type: Boolean,
-      default: true
-    },
```
NicoPennec, line 60: This useless property is still used in `Sequences.vue` and `Shots.vue`.

NicoPennec (approved):

PR #1676, opened 2025-01-22
various improvements

EvanBldy:
**Problem**
- When creating a playlist on the fly, it sends an event "preview-room:*" with "temp" as room_id, causing an error in zou-events.
- In review room, the laser mode is not working. 

**Solution**
- Stop sending that event when room.id is not a correct id. 
- Fix laster mode in review room, it was due to the fact that addObjectToCanvas is now a promise.

### DIFF
--- a/src/components/mixins/previewRoom.js
+++ b/src/components/mixins/previewRoom.js
@@ -273,8 +273,7 @@ export const previewRoomMixin = {
         const obj = eventData.data.obj
         if (this.getObjectById(obj)) return
         if (this.isLaserModeOn) {
-          const o = this.addObjectToCanvas(annotation, obj)
-          this.fadeObject(o)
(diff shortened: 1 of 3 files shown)


src/components/mixins/previewRoom.js
```diff
@@ -273,8 +273,7 @@ export const previewRoomMixin = {
         const obj = eventData.data.obj
         if (this.getObjectById(obj)) return
         if (this.isLaserModeOn) {
-          const o = this.addObjectToCanvas(annotation, obj)
-          this.fadeObject(o)
+          this.addObjectToCanvas(annotation, obj).then(o => this.fadeObject(o))
```
frankrousseau, line 276: Are you sure that the object is not saved when you add this?

PR #1699, opened 2025-02-18
Various improvements

frankrousseau:
**Problem**

* It's not possible to see assignees in the playlist
* Annotations are not saved when the task panel is closed too quickly
* It's not possible to add the daily pending to asset playlists

**Solution**

* Display assignees in the comment panel in the playlist for supervisors and managers
* Use task ID from the preview file if the task information is no longer available.
* Add a "Add daily" pending button to the asset playlist creator.

### DIFF
--- a/src/store/modules/assets.js
+++ b/src/store/modules/assets.js
@@ -1,3 +1,5 @@
+import moment from 'moment'
+
 import assetsApi from '@/store/api/assets'
 import peopleApi from '@/store/api/people'
 import assetTypeStore from '@/store/modules/assettypes'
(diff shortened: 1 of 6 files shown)


src/store/modules/assets.js
```diff
@@ -847,6 +849,31 @@ const actions = {
...
+      let isPending = false
+      asset.tasks.forEach(taskId => {
+        const task = tasksStore.state.taskMap.get(taskId)
+        if (!isPending) {
+          if (daily) {
+            if (task.last_comment_date) {
+              const lastCommentDate = moment(task.last_comment_date)
+              const yesterday = moment().subtract(1, 'days')
+              isPending =
+                task.task_status_short_name === 'wfa' &&
```
NicoPennec, line 866: Testing an editable task status name seems unreliable.

Why not test on the "is_feedback_request" property of the task status?

frankrousseau: Thx, it's fixed

Issue #1673, opened 2025-01-21
Quota Weekly download CSV missing data


scottmcdonnell:
**Context**

Studio name:  JAM
Kitsu version: 0.20.19

**Describe the bug**

In the quotas section the Daily figures download as a CSV correctly but if you switch to **Detail Level: Week** and any Count Unit or Compute Mode -- the downloaded CSV is missing all values.
It works correctly for 2024 but does NOT for 2025. 

This is different to #1612 

**Screenshots**

![Image](https://github.com/user-attachments/assets/b3ef2362-cd84-4d21-b979-de81f0600f25)

scottmcdonnell:
it looks like the zou api side. The request - quota/week/2025?countMode=seconds returns the weeks without padding:
2025-1 should be 2025-01 for the getQuotaEntries() to work.

![Image](https://github.com/user-attachments/assets/b102e07f-4f3d-4bf1-a6bd-fa3f97365a1d)

scottmcdonnell:
Add fix for day quota download bug - #1713

scottmcdonnell:
closing as fix merged

PR #1702, opened 2025-02-19
bugfix for download quotas csv for weekly figures returning empty csv

scottmcdonnell:
**Problem**
CSV download for weekly quotas was giving empty results for weeks early in the year. see issue #1673


frankrousseau:
Thank you @scottmcdonnell!

PR #1710, opened 2025-02-25
Tasks improvements

frankrousseau:
**Problem**

* It's not possible to filter the All Tasks page by studio or by department
* There is no task information in the task schedule page
* There is no way to manage 2D productions made with papers

**Solution**

* Add studio, and department filters to the all tasks page
* Add information from the task list to the schedule page
* Add drawings statistics for production that have the new style 2D Paper


src/components/cells/TimeSliderCell.vue
```diff
@@ -10,7 +10,7 @@
         :interval="0.25"
         :lazy="true"
         :min="0"
-        :max="maxHoursByDay"
+        :max="12"
```
NicoPennec, line 13: Why change this max value?
A bigger "hoursByDay" value can be selectable with the button on the right of slider (see line Line 25)

src/locales/en.js
```diff
@@ -619,10 +620,12 @@ export default {
     more_filters: 'More filters',
     move_action_bar: 'Move action bar',
     no_estimation: 'No estimation',
+    nb_drawings: 'drawing | drawings',
```
NicoPennec, line 623: duplicate value (line 594 / line 623)

src/components/pages/ProductionQuota.vue
```diff
@@ -233,7 +311,7 @@ export default {
     getCurrentPerson() {
       const personId = this.$route.params.person_id
       if (personId && this.personMap) {
-        return this.personMap.get(personId)
+        return personMap.get(personId)
```
NicoPennec, line 314: Inconsistent test (`this.personMap`) vs usage (`personMap`)

src/components/widgets/PeopleField.vue
```diff
@@ -55,15 +62,22 @@ export default {
...
-    this.item = this.modelValue
     this.index = buildNameIndex(this.people)
   },
 
   mounted() {
     this.items = this.people
+    this.item = this.modelValue
+    setTimeout(() => {
+      this.item = this.modelValue
+    }, 10)
```
NicoPennec, line 73: Why this code?

src/components/cells/TimeSliderCell.vue
frankrousseau: It's because it breaks the styling.

src/components/pages/ProductionQuota.vue
frankrousseau: I don't understand your comment.

src/components/widgets/PeopleField.vue
frankrousseau: Because the item is not properly selected on mount. The value is right but the person name doesn't appear in the widget.

PR #1713, opened 2025-02-25
bugfix for daily quotas giving the incorrect date range for days.

scottmcdonnell:
**Problem**
download the daily quotas for month January 2025 and the headers are for the current month, February.

**Solution**
There was a mix up between month and currentMonth, and todayMonth.  Changed to todayMonth to make it consistent accross the csv.js file.


frankrousseau:
Thank you @scottmcdonnell for your contribution!

PR #1731, opened 2025-03-28
Various fixes

frankrousseau:
**Problem**

* When the first page of a production is set as "shots" and the production is a tv series, the section combobox is not updated properly
* Pagination system in the todo list page leads to unexpected side effects in the time sheet section (some tasks are missing)

Edit:

* The department selection is lost when browsing from one page to another.
* There is room for bigger thumbnails in the playlist list.
* The playlist navigation ignores the sub previews when using the next and previous buttons.

**Solution**

* After a production change is noticed and applied in the topbar section, always sync topbar combo box values with the route parameters.
* Remove the pagination system from the todo-list (not really useful).

Edit:

* Keep the last selection if it is available in the combo box (do not look for departments to which the person belongs).
* In the playlist list, make thumbnails bigger.
* When using the next and previous buttons, it jumps to the following sub preview before jumping to the following entity.

### DIFF
--- a/src/components/lists/TodosList.vue
+++ b/src/components/lists/TodosList.vue
@@ -307,7 +307,6 @@ import { selectionListMixin } from '@/components/mixins/selection'
 import { formatListMixin } from '@/components/mixins/format'
 import { descriptorMixin } from '@/components/mixins/descriptors'
 
-import { PAGE_SIZE } from '@/lib/pagination'
 import { sortPeople } from '@/lib/sorting'
 import {
   daysToMinutes,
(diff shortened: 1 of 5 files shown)


src/components/tops/ActionPanel.vue
```diff
@@ -304,7 +304,7 @@
         </div>
 
         <div
-          class="menu-item mr05"
+          class="menu-item mr05 cursor"
```
NicoPennec, line 307: The `cursor ' class doesn't exist. Do you mean `pointer`?

```suggestion
          class="menu-item mr05 pointer"
```

NicoPennec (approved):

PR #1748, opened 2025-04-29
[Budget] Various fixes

frankrousseau:
**Problem**

* The icon style is wrong
* Null value is not properly handled in the edit person modal
* People are not filtered by department in the edit budget entry modal
* When someone has no seniority or position, budget entry addition is not properly handled
* Department collapsing is not saved
* Monthly chart displays only values over 12 months
* Adding an entry after an edition doesn't reset the modal

**Solution**

* Make the outline width cohesive
* Add a null choice to the related combo box
* Filter the  people list on department change
* Use default values when seniority or position is missing
* Save the  department collapsing state in the local storage
* Set labels properly to avoid using the same one for different months
* Reset the entry to edit before opening the add modal

### DIFF
--- a/src/components/tops/TopbarSectionList.vue
+++ b/src/components/tops/TopbarSectionList.vue
@@ -15,6 +15,7 @@
           />
           <hand-coins-icon
             class="section-icon"
+            :stroke-width="1.5"
             v-if="currentSectionValue === 'budget'"
           />
           {{ currentSectionLabel }}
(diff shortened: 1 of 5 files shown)


src/components/tops/TopbarSectionList.vue
```diff
@@ -39,6 +39,7 @@
             />
             <hand-coins-icon
               class="section-icon"
+              :stroke-width="1.5"
```
NicoPennec, line 42: Maybe add the `:stroke-width` to line the Line 17 too?
There is the samed icon component.

NicoPennec (approved):

PR #1750, opened 2025-04-29
Add title on rowActionsCell icons

lechristianfillon:
**Problem**
There are no tooltips associated with the icons in the action column on rows in a table. 
The title property is not defined on icons.

**Solution**
Creation and translation of labels in the src/locales directory. Implementation of the "title" properties on the icons in the RowActionsCell.vue file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kitsu",
-  "version": "0.20.52",
+  "version": "0.20.54",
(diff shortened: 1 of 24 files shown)


frankrousseau:
No problem. It's just that it will add noise if you keep the locale files. Can you remove them?

frankrousseau:
Thank you for the modifications!

PR #1753, opened 2025-05-02
Various fixes

frankrousseau:
**Problem**

* The unassign all button doesn't work anymore
* The asset page reloads every time a tab is changed
* The playlist raises an error message related to the waveform widget
* There is no total line in the entity page about task information

**Solution**

* Add the missing variable
* Fix the reloading condition and use the entity mixin
* Loadthe  waveform only if it is displayed
* Add a total line to the entity task list in the entity page

### DIFF
--- a/src/components/lists/EntityTaskList.vue
+++ b/src/components/lists/EntityTaskList.vue
@@ -98,6 +98,18 @@
             </td>
             <td class="end-cell"></td>
           </tr>
+          <tr class="datatable-row total-row">
+            <td>{{ $t('main.total') }}</td>
+            <td>{{ entityProgress }}</td>
+            <td class="estimation">{{ formatDuration(entityEstimation) }}</td>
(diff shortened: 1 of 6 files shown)


src/components/lists/EntityTaskList.vue
```diff
@@ -98,6 +98,18 @@
             </td>
             <td class="end-cell"></td>
           </tr>
+          <tr class="datatable-row total-row">
+            <td>{{ $t('main.total') }}</td>
+            <td>{{ entityProgress }}</td>
+            <td class="estimation">{{ formatDuration(entityEstimation) }}</td>
+            <td class="estimation">{{ formatDuration(entityDuration) }}</td>
```
NicoPennec, line 105: Wrong class name:
```suggestion
            <td class="duration">{{ formatDuration(entityDuration) }}</td>
```

NicoPennec (approved):

PR #1754, opened 2025-05-05
Peoplelisteresizing

lechristianfillon:
**Problem**
In screen "People", new columns have been added to the people table without resizing existing columns.

**Solution**
Column widths have been optimized to have a default table width that displays the actions column without requiring a horizontal scrollbar. 
This depends on the window size. The default behavior is for a 1920x1080 window.


NicoPennec:
Thank you @lechristianfillon for your contribution 🙏 

I just made some adjustments to a few minimum sizes.

PR #1758, opened 2025-05-11
Modification description issue by modals component

lechristianfillon:
**Problem**
On entities list page (asset, edit, episode, sequence, shot), description cell not dynamically updated after description modification through the modal pop-up opened by the edit button action.
On Asset page, bottom description cell is not not dynamically updated after description modification through the modal pop-up opened by the edit button.
Javascript error on Episode page if we try to modify the episode information by through the modal pop-up just after the episode creation (without navigation to refresh the page)

**Solution**
In store mutation, broadcasts the modified object into the array of displayed objects then used in the corresponding computed variable.
Remove dead code

### DIFF
--- a/src/components/pages/Asset.vue
+++ b/src/components/pages/Asset.vue
@@ -634,7 +634,6 @@ export default {
     getCurrentAsset() {
       return new Promise((resolve, reject) => {
         const assetId = this.route.params.asset_id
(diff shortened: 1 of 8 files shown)


NicoPennec (approved):
Thanks @lechristianfillon for your helpful correction.

Issue #1764, opened 2025-05-15
DescriptionCell invisible on the first table row in the Sequences and Episodes entity management page


lechristianfillon:
**Context**

Studio name: Personal project
Kitsu version: 0.20.57

**Describe the bug**

The DescriptionCell component is designed to be displayed above the selected cell. 
This works in most entity management screens because there is sufficient space above the first table, which is included in a global table.
In the case of the Sequences and Episodes entities, there is only one table with the header at the very top of the page. There is no space to display the DescriptionCell, the description text is not visible, and editing is impossible because the cursor is also invisible.

This is due to the ".datatable-wrapper { overflow: auto;" css property, which would be risky to change.

Below are screenshots for both entities, with the DescriptionCell open but hidden.

**Screenshots**

![Image](https://github.com/user-attachments/assets/64725f7e-a73c-4322-aad4-6fbbdf0eb7d4)

![Image](https://github.com/user-attachments/assets/27a3633c-db8e-405d-abf8-d618cbff5b57)

**Solution**

There are several possible solutions, some more complex than others, and some with varying degrees of risk for the entire application.
I propose this first solution with only a small, risk-free CSS modification: display DescriptionCell to the right of the cell, on the currently selected row.
If you accept it, I can deliver it.
Otherwise, I'll let you make another proposal and see if I can take care of it.

Below is the result.
![Image](https://github.com/user-attachments/assets/775c4edf-4386-460f-8f30-ca0bbec34b31)

 - OS: Windows
 - Browser: Brave
 - Version: 1.78.94 Chromium: 136.0.7103.60 (Build officiel)

NicoPennec:
Hi @lechristianfillon,

It's a known UI problem. We have a similar issue with the combobox: https://github.com/cgwire/kitsu/issues/855

CSS solutions do not fully address scroll overflow issues, which can occur from the bottom or sides. We've already tried this solution.

Since our recent move to Vue 3, a new solid alternative is to use a `<teleport>` view to place the pop-in at the top level of the DOM.
Our date field, based on `@vuepic/vue-datepicker`, uses this new solution.

lechristianfillon:
Ok, good option. Done.

NicoPennec:
Fixed on Kitsu >= 0.20.58

Thanks for your wonderful help @lechristianfillon 🙏

PR #1765, opened 2025-05-18
Implement teleport for description cell tooltip

lechristianfillon:
**Problem**

DescriptionCell tooltip is invisible on the first table row in the Sequences and Episodes entity management page.
Full description in related issue: https://github.com/cgwire/kitsu/issues/1764

**Solution**
NicoPennec suggested using the Vue3 feature "teleport" => good option. It is implemented in this PR.

### DIFF
--- a/src/components/cells/DescriptionCell.vue
+++ b/src/components/cells/DescriptionCell.vue
@@ -17,28 +17,32 @@
         "
       >
       </span>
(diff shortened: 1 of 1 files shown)


NicoPennec:
Nice work @lechristianfillon !
Another helpful contribution 🙏

Issue #1576, opened 2024-11-05
If the title of a sequence contains several words, the filter doesn't work


gwendupre:
**Context**

Studio name: gwen
Kitsu version:  0.19.72
Production type: all

**Describe the bug**

In the shot page, clicking on a sequence name filters the page with that sequence. The result is identical if you type the sequence name in the filter bar.
However, if the sequence title is made up of several words, the filter doesn't work.
Example: if a sequence is called “blue rabbit”, clicking on the name of the sequence won't return any results, and neither will typing rabbit nor blue in the filter bar.

**Desktop (please complete the following information):**

 - OS: Windows
 - Browser: Chrome
 - Version 130.0.6723.92 (Build officiel) (64 bits)

PR #1772, opened 2025-05-26
Allow searching for shots by name with name including space

lechristianfillon:
**Problem**
https://github.com/cgwire/kitsu/issues/1576
if a sequence is called “blue rabbit”, clicking on the name of the sequence won't return any results, and neither will typing rabbit nor blue in the filter bar.

**Solution**
Main constraint: spaces are used as a separator to manage different filter criteria.

To avoid changing this feature and allow names containing spaces, we can add " around the names to protect the space contained in the name.
The characters " are added automatically when you click on the sequence name. You can also add them manually in the search field.

Examples available after this modification:
    "lapin bleu" [Storyboard]=[todo] assignedto[Storyboard]=[john Doe]
    "nom plusieurs mots"
    [Storyboard]=[todo] aa"aa

### DIFF
--- a/src/lib/filtering.js
+++ b/src/lib/filtering.js
@@ -191,9 +191,9 @@ const applyFiltersFunctions = {
 /**
  * Extract keywords from a given text. Remove equality and exclusion
  * expressions.
- * Replace spaces inside quotes with '@' to allow searching for names containing spaces
+ * Replace spaces inside quotes with '\u00A0' to allow searching for names containing spaces
  *    (e.g., "my sequence" becomes "my@sequence")
- * The '@' will be removed after splitting on spaces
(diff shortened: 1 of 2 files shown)


src/lib/filtering.js
```diff
@@ -191,18 +191,36 @@ const applyFiltersFunctions = {
 /**
  * Extract keywords from a given text. Remove equality and exclusion
  * expressions.
+ * Replace spaces inside quotes with '\u00A0' to allow searching for names containing spaces
+ *    (e.g., "my sequence" becomes "my@sequence")
```
NicoPennec, line 195:
```suggestion
 *    (e.g., "my sequence" becomes "my\u00A0sequence")
```

PR #1775, opened 2025-05-27
Various fixes

frankrousseau:
**Problem**

* Budget real costs don't include unlisted users
* The breakdown shows the asset type section from the library while they should be hidden
* The on-the-fly playlist progress is bugged

**Solution**

* Add unlisted users to the budget when expenses are displayed.
* Do not include library assets when checking if a type should be hidden
* Add duration information to previews

### DIFF
--- a/src/components/pages/budget/BudgetList.vue
+++ b/src/components/pages/budget/BudgetList.vue
@@ -502,6 +502,11 @@ export default {
       'organisation'
     ]),
 
+    /* It converts the expenses to the budget format where there is an
+     * entry for each department and each person. It also calculates the totals
+     * for each person, department and for all the departments.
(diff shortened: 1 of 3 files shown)


src/components/pages/budget/BudgetList.vue
```diff
@@ -540,11 +545,85 @@ export default {
...
+            acc[department.id][entry.person_id] = true
+          })
+          return acc
+        },
+        {}
+      )
+
+      Object.keys(this.expenses).forEach(departmentId => {
+        if (departmentId === 'total') return
+        console.log(departmentId)
```
NicoPennec, line 594: remove console.log
```suggestion
```

PR #1776, opened 2025-05-28
[budget] Improve expenses styles

frankrousseau:
**Problem**

* In the budget section, the expenses are hard to read.

**Solution**

* Show only expenses, not the estimation for concerned months. 
* Add colors to expenses headers
* Put expense aggregates after the expense columns

### DIFF
--- a/src/components/pages/budget/Budget.vue
+++ b/src/components/pages/budget/Budget.vue
@@ -29,6 +29,8 @@
           :is-error="errors.entries"
           :is-loading="loading.entries"
           :is-showing-expenses="expenses.showing"
+          :months-between-start-and-now="monthsBetweenStartAndNow"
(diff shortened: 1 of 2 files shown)


src/components/pages/budget/Budget.vue
```diff
@@ -649,9 +664,8 @@ export default {
       const months = []
       const current = parseSimpleDate(startDate)
       const end = parseSimpleDate(endDate)
-
       while (current <= end) {
-        months.push(moment(current))
+        months.push(moment(current.clone()))
```
NicoPennec, line 668: Why not just `clone` here?

```js
months.push(current.clone())
```

frankrousseau: thx, I fixed that!

Issue #712, opened 2021-06-09
Filters do not support accents


angel-k:
On Kitsu 0.12.81
When I try to filter by Assignation with an user who has an accent in their name, the filter gives me 0 results. It seems that the filters do not handle accents.
I have the same problem when a Task Type have an accent.

Example :
The user is called "Léa" and is assigned to tasks.
I search with the filters :
- assignedto=[Léa] -> 0 result
- assignedto=[Lea] -> doesn't filter anything

frankrousseau:
Thank you for reporting!

frankrousseau:
It should be fixed. Can you try again?

angel-k:
I didn't test it before today (Kitsu 0.13.0)
The accents support for Task Type and Assignation seems to work well but now I feel like the "assignedto" filter is buggy...
For some users (with or without accents) it works, for others I get 0 results. It seems to happen when multiple user have the same name (or almost the same). It worked well before.
I guess that the accent support is fixed but that the "assignedto" filter not working well is a new bug ?

frankrousseau:
Mmm it's embarassing. That bug happened before but if you have two artists with the exact same name, I'm not sure that we can fix this.

frankrousseau:
And without concrete examples, I won't be able to fix this issue

angel-k:
Here's an example:

If I create 2 users Clément Deschamps and I assign them both a task. When I search for the tasks assigned to Clément Deschamps, only those of the 2nd user are displayed.
If I change the name of the 2nd Clément by Clément DESCHAMPS. When I search for the tasks assigned to Clément Deschamps, same situation, only those of the 2nd Clément are displayed.
![case1](https://user-images.githubusercontent.com/25372475/127019723-abade022-4bc7-471c-a5c8-291e73376038.png)

If I change the name of the 2nd Clement to Clement DESCHAMPS. When I search for the tasks assigned to Clement Deschamps, only those of the 1st Clément are displayed.
I notice that when I renamed the 2nd Clément to Clement DESCHAMPS, he changed position in the user list.
![case2](https://user-images.githubusercontent.com/25372475/127019743-a43240a8-72aa-4f14-8d57-7080bb0cb601.png)

If I disable the 2nd Clément, I revert to the initial situation, only the tasks of the 2nd Clément are displayed.
![case3](https://user-images.githubusercontent.com/25372475/127019951-921bfc8f-81da-49fc-93eb-a071e6208b2e.png)

[...]

PR #1778, opened 2025-05-28
Allows filtering of duplicate names in search on Assets page

lechristianfillon:
**Problem**
Related to issue: https://github.com/cgwire/kitsu/issues/712
First issue was solved.
Second issue is about "multiple users have the same name" and was not solved.

If several persons have the same name (first and last name), only one person is considered by filtering.

**Solution**
There were two issues with the filter handling for the person assigned to the task:
- a bug: in a forEach statement, "return true / return false" was coded to exit the loop, which is incorrect because the loop continues to move on to the next item. The function has been replaced with a simple for loop allowing the loop to be exited with a simple break. 

- the suggestion to use the person's ID was correct and already present in the code, transparent to the user. However, there was a flaw in the code because a map was used on the "Name" criterion, which automatically removed elements with the same key. The map is replaced with a simple array of objects, which allows the addition of people with the same name. Searching the array then finds all occurrences of people equal to the name specified in the filter.

In the screenshot below, I demonstrate the use of a complex filter, allowing to find the tasks associated with a username corresponding to 2 different people, as well as the tasks associated with another username

image

### DIFF
--- a/src/lib/filtering.js
+++ b/src/lib/filtering.js
@@ -4,7 +4,6 @@ import {
   buildNameIndex,
   indexSearch
 } from '@/lib/indexing'
-import string from '@/lib/string'
 
 const UNION_REGEX = /\+\(.*\)/
 const EQUAL_REGEX =
@@ -41,6 +40,15 @@ export const applyFilters = (entries, filters, taskMap) => {
   } else {
     return entries
   }
+}
+
+/*
+ * Private function for filtering on names
+ * Simple hash to compare user name
+ * It removes spaces and lowercases the name
+ */
+const hashName = name => {
+  return name.toLowerCase().replace(/ /g, '')
 }
 
 const applyFiltersFunctions = {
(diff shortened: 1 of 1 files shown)


NicoPennec (changes requested):
Hi @lechristianfillon 

Thank you for the proposal.
After some testing, this seems incorrect to me.
It doesn't work with two users with the same name and a simple filter.

Example:
- task A assigned to "clement"
- task B assigned to "clément"
- task C assigned to "clement" and "clément"

Filter:
`assignedto=[clément]`

Result:
- task C assigned to "clement" and "clément"

The task B is missing from result.

lechristianfillon:
Check with:
assignedto[Modeling]=[clement unknown]
assignedto[Modeling]=[clement unknown]
assignedto[Modeling]=[clément unknown] assignedto[Modeling]=[clement unknown]
+(assignedto[Modeling]=[clément unknown] assignedto[Modeling]=[clement unknown])

assignedto[]=[clément unknown]
assignedto[]=[clement unknown]
assignedto[]=[clement unknown] assignedto[]=[clément unknown]
+(assignedto[]=[clement unknown] assignedto[]=[clément unknown])

Issue #1531, opened 2024-09-16
Breakdown columns alignment


frankrousseau:
**Context**

Studio name: Ryff
Kitsu version: 0.19.56
Production type: Short

**Describe the bug**

Breakdown metadata columns are not properly aligned

**Screenshots**

![image](https://github.com/user-attachments/assets/14d12716-c7c8-44bb-9bf6-26a989c0d63d)

frankrousseau:
It was fixed

PeteDraper:
sorry, still having same / similar issue in 0.20.5 - possibly this is due to the size of the custom fields 
![Screenshot 2024-12-18 at 20 00 39](https://github.com/user-attachments/assets/e9d8db9c-2081-487b-829c-bb8f7461ac92)

PeteDraper:
where the text isn't so big, it seems to be ok? or could it be due to the fact that the 1st screenshot is from a film setup and the second from a tv-show ? puzzling..

![Screenshot 2024-12-18 at 20 04 18](https://github.com/user-attachments/assets/d10d6ec1-2bc8-458d-8b45-69cb2315efc6)

PeteDraper:
it seems like the slider between the NAME and STDB columns is unnaturally long. If it's brought to the left and aligned it matches. Refreshing the page it keeps the position, correctly aligning the columns. However the large text fields are not being split over a 2 line title height. Additionally, the ability to resize the columns is limited to 13, including Name and STDB

PR #1779, opened 2025-05-29
width style fix for breakdown columns alignment

lechristianfillon:
**Problem**
Related to : https://github.com/cgwire/kitsu/issues/1531

Some columns have width issues between header and lines

**Solution**
2 issues
Metadata columns
image
Solution => change from 119px to 110px for metadata header width

Name column
When the page is displayed in a smaller window
image
Solution => add min-witdh on name line cell


NicoPennec:
Thank you, @lechristianfillon, for the fix!

PR #1802, opened 2025-06-30
Feature: Comment hashtag for task types

scottmcdonnell:
This PR adds a new feature to the comments to allow users to use a hashtag to target comments to another task type on the same entity.  This is similar to tagging users with the "@" sign, you can type "#" to bring up the list of task types for this entity.

For example in the "Animation" task for a given shot, the artist can send a comment to "Final Comp" 
image

You can hashtag several task types in one comment and once rendered they will appear with a link that takes you to the task.
This is how the comment now appears on the task you added the comment to:
image

"#All" will send a comment to all tasks for the entity. 

and on each task that you added a hashtag there will be a comment added.  
The status is not changed by the added comment. 
The asterisk will show where the comment came from. 
image

A matching Zou PR is required for this feature, in order to save the comment.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8681,7 +8681,8 @@
     },
     "node_modules/vue-at": {
       "version": "3.0.0-alpha.3",
-      "resolved": "git+ssh://git@github.com/scottmcdonnell/vue-at.git#8edaa534f960a3b79ce668c054236acd332fbb83",
+      "resolved": "https://registry.npmjs.org/vue-at/-/vue-at-3.0.0-alpha.3.tgz",
+      "integrity": "sha512-ScbPjh29IbqJvWSqcDci6VW5q3xnbk/1axnIBjZkQjDe4HZzWHuzwjJZQEksxn/vK84AruiTdiW/ESf4RJiHyg==",
       "dependencies": {
         "textarea-caret": "^3.1.0"
       },


scottmcdonnell:
There is a zou PR required for this feature, my colleague is cleaning it up and will add it shortly.

frankrousseau:
We can't depend from one of your repositiory. What did you change in `vue-at`?

scottmcdonnell:
> vue-at

Yes I had to send the member object to the filtermatch to allow the filtering to work here - I added a PR https://github.com/fritx/vue-at/pull/152

frankrousseau:
Thank you for the details. 

It seems that At is longer maintained. We should explore a migration to: https://vue-tribute.netlify.app/

src/components/modals/EditCommentModal.vue
```diff
@@ -370,18 +399,43 @@ export default {
...
       deep: true,
       immediate: true,
       handler() {
+        let teamOptions = []
         if (this.isCurrentUserClient) {
-          this.atOptions = this.team.filter(person =>
-            ['admin', 'manager', 'supervisor', 'client'].includes(person.role)
-          )
+          teamOptions = [
+            ...this.team.filter(person =>
```
NicoPennec, line 431: Why add a spread operator?
It seems pointless to me because `filter` returns a shallow copy.

scottmcdonnell:
I can on return from annual leave 👍

PR #1808, opened 2025-07-03
adjust Asset and Shot page task-list and info scrolling behaviour to …

scottmcdonnell:
**Problem**
The entity pages for Shot and Asset have the task-list and info as scroll areas. The are not consistent with each other and the scroll areas make readability difficult.

**Shot entity page**
problems in screenshot:
- 3 scrollbars
- table cut off when scrolling right
- information not appearing

image

**Asset entity page**
problems in screenshot:
- hard to see scrollbars
- hard to see there are more tasks
image

**Solution**
Make one large scroll area for task-list and information consistent across both Asset and Shot entities.
image

### DIFF
--- a/src/components/pages/Asset.vue
+++ b/src/components/pages/Asset.vue
@@ -105,7 +105,7 @@
             </div>
           </div>
 
-          <div class="table-body metadata-infos">
+          <div class="table-body entity-infos">
             <table class="datatable no-header" v-if="currentAsset">
               <tbody class="table-body">
                 <tr class="datatable-row">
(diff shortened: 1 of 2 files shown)


frankrousseau:
Hello @scottmcdonnell. It doesn't match the behaviour to expect: separate scrolls for both lists. Can you tell us more about what is not working for you?

frankrousseau:
We were missing information to merge this one.

scottmcdonnell:
Sorry, I have been on leave. The Asset page in particular has multiple scroll areas as per my screen shots in the first post. For us the task list gets cut off and gets very small when the browser is not full screen, if you decrease the browser window size the task list disappears. This is different to other entity pages, shot and asset both benefit from a single scroll area for tasks and information rather than separate scroll areas for each.

PR #1809, opened 2025-07-03
In dark theme fix the scrollbar colour over a block area

scottmcdonnell:
**Problem**
Scrollbar was not visible when a scroll area was within a div with the "block" class as the scrollbar colour was the same colour as the block area.

**Solution**
css update

image


frankrousseau:
Thank you!

PR #1815, opened 2025-07-08
[csv][estimations] hide csv import button for task type estimations

EvanBldy:
**Problem**
- The csv import button for task type estimations is always shown. 

**Solution**
- Hide import button when importing task type estimations is not allowed due to permissions.

### DIFF
--- a/src/components/pages/TaskType.vue
+++ b/src/components/pages/TaskType.vue
@@ -22,12 +22,18 @@
               @click="contactSheetMode = !contactSheetMode"
               v-if="isActiveTab('tasks')"
             />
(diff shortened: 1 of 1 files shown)


src/components/pages/TaskType.vue
```diff
@@ -639,6 +645,7 @@ export default {
       const departments = this.user.departments || []
       return (
         this.isCurrentUserManager ||
+        !departments.length ||
```
NicoPennec, line 648: Rule to fix:

`this.isCurrentUserSupervisor && ( !departments.length || departments.includes(...) )`

PR #1846, opened 2025-08-26
Show the hard delete modal for tasks, assets, shots, edits, concepts to avoid accidental data loss.

scottmcdonnell:
**Problem**
We had a user with correct permissions accidentally delete a number of tasks when trying to create a playlist.

There is a hard delete prompt message when deleting episodes, sequences and budgets to let the user know that the delete button will permanently lose the data. 
When you delete tasks they hard delete with no warning losing all comments, previews etc. 
When you delete entities such as shots and assets they soft delete first (canceled = True) then the second time they hard delete with no warning. 
Concepts delete first time, with no soft delete option. 

**Solution**
This PR adds a hard delete modal for these extra cases where data will be lost on pressing the delete button and differentiates between the soft delete. This makes for a consistent experience for users and helps avoid accidental deletion.

### TASKS
Select tasks and delete should always show a hard delete modal as tasks does not have a soft delete option. 

image

Type DELETE to approve the permanent deletion. 

image

### ASSETS, EDITS, SHOTS - soft delete
Cancelled entities now show as strike-through in the list to differentiate 

image

Assets, edits and shots are soft deleted if any selected item is not already cancelled. The list above will be soft deleted and no hard delete modal will appear on clicking delete button.

### ASSETS, EDITS, SHOTS - hard delete

We check if all items are cancelled already and if so the process will be a hard deleted.

image

For this list all items are already cancelled so the hard delete modal will show.

image

### Concepts
Although concepts have a soft delete, UI currently hard deletes first time. Always on hard delete modal added.

### DIFF
--- a/src/components/sides/TaskInfo.vue
+++ b/src/components/sides/TaskInfo.vue
@@ -305,7 +305,9 @@
             :key="entity.id"
             v-for="entity in Array.from(selectedEntities.values())"
           >
-            <span :class="{ canceled: entity.canceled }">{{ entity.full_name }}</span>
+            <span :class="{ canceled: entity.canceled }">{{
+              entity.full_name
+            }}</span>
           </div>
         </div>
       </div>
--- a/src/components/sides/TaskInfo.vue
+++ b/src/components/sides/TaskInfo.vue
@@ -305,9 +305,7 @@
             :key="entity.id"
             v-for="entity in Array.from(selectedEntities.values())"
           >
-            <span :class="entity.canceled ? 'canceled' : ''">{{
-              entity.full_name
-            }}</span>
+            <span :class="{ canceled: entity.canceled }">{{ entity.full_name }}</span>
           </div>
         </div>
       </div>
--- a/src/components/sides/TaskInfo.vue
+++ b/src/components/sides/TaskInfo.vue
@@ -305,7 +305,9 @@
             :key="entity.id"
             v-for="entity in Array.from(selectedEntities.values())"
           >
-            {{ entity.full_name }}
+            <span :class="entity.canceled ? 'canceled' : ''">{{
+              entity.full_name
+            }}</span>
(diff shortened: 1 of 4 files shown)


src/components/tops/actions/DeleteEntities.vue
```diff
@@ -12,16 +12,29 @@
...
+    :active="modals.deleteConfirmation"
+    :error-text="errorText"
+    :is-loading="isLoading"
+    :is-error="isError"
+    :text="hardDeleteTextComputed"
+    :lock-text="hardDeleteLockTextComputed"
+    @cancel="modals.deleteConfirmation = false"
+    @confirm="confirm"
+    vif="modals.deleteConfirmation"
+  />
```
NicoPennec, line 25: - Code typo on `v-if`
- `active` can always be set to `true` by default if a `v-if` is used.

```suggestion
  <hard-delete-modal
    active
    :error-text="errorText"
    :is-loading="isLoading"
    :is-error="isError"
    :text="hardDeleteTextComputed"
    :lock-text="hardDeleteLockTextComputed"
    @cancel="modals.deleteConfirmation = false"
    @confirm="confirm"
    v-if="modals.deleteConfirmation"
  />
```

src/components/sides/TaskInfo.vue
```diff
@@ -305,7 +305,9 @@
             :key="entity.id"
             v-for="entity in Array.from(selectedEntities.values())"
           >
-            {{ entity.full_name }}
+            <span :class="entity.canceled ? 'canceled' : ''">{{
+              entity.full_name
+            }}</span>
```
NicoPennec, line 310: You can replace ternary with object syntax for class binding.

```suggestion
            <span :class="{ canceled: entity.canceled }">{{ entity.full_name }}</span>
```

PR #1881, opened 2025-10-16
[playlists] Allow to add several time the same entity

frankrousseau:
**Problem**

It's not possible to review several versions of the same entity in the same playlist.

**Solution**

* Identfiy playlist elements via the couple `entity_id` - `preview_file_id`. 
* Do not allow the same preview to be put twice.
* Remove callbacks and replace them with promises in the playlist store.

### DIFF
--- a/src/components/pages/playlists/PlaylistPlayer.vue
+++ b/src/components/pages/playlists/PlaylistPlayer.vue
@@ -2264,7 +2264,6 @@ export default {
     resetPlaylistFrameData() {
       let playlistDuration = 0
       let currentFrame = 0
-      console.log('resetPlaylistFrameData', this.entityList)
       this.entityList.forEach((entity, index) => {
         const defaultNbFrames =
           entity.preview_nb_frames || 2 * this.fps * this.frameDuration


src/components/pages/Playlist.vue
```diff
@@ -1013,15 +1022,20 @@ export default {
     },
 
     addEntityToPlaylist(entity) {
-      if (this.currentEntities[entity.id]) {
+      /*if (this.currentEntities[entity.id]) {
         return
-      }
+      }*/
```
NicoPennec: comment to remove?

src/store/modules/playlists.js
```diff
@@ -23,23 +23,49 @@ import {
...
+      previewFile = previewFiles.shift()
+      const key = `${entity.id}-${previewFile.id}`
+      if (!state.playlistEntryMap.has(key)) {
+        state.playlistEntryMap.set(key, previewFile)
+        found = true
+      }
+    }
+    if (!found) {
+      previewFile = null
+    }
```
NicoPennec: Using `previewFiles.find()` will be more efficient and readable, no?

```js
  const previewFile = previewFiles.find(file => 
    !state.playlistEntryMap.has(`${entity.id}-${file.id}`)
  );

  if (previewFile) {
    state.playlistEntryMap.set(`${entity.id}-${previewFile.id}`, previewFile);
  }
```

PR #1891, opened 2025-10-28
Update vite version to ^5.0.0

Antonytj:
Updated vite dependency version in package.json.

**Problem**

**Solution**


NicoPennec:
We are on Vite 7...

PR #1938, opened 2026-01-26
bugfix: assignment silently failing 

scottmcdonnell:
**Problem**
When a Supervisor assigns a person not in their department to a task it silently fails. The API returns an empty list [] for the `api/actions/persons/<id>/assign` call, but the UI updates the assignee for the task and it looks to the user like that task has been successfully assigned. On page reload the assignment "disappears".

Similarly if a Supervisor selects a task in their department, then more tasks including those NOT in their department and assigns them to a user in their team, then the api returns a list of tasks successfully assigned omitting those not assigned for whatever reason. Again the UI shows ALL selected tasks assigned instead of just the successfully assigned and a page reload fixes this. 

**Solution**
Get the assigned task ids back from the api response and only update the tasks for those. Show an error if any failed assignments. 
image

### DIFF
--- a/src/store/modules/tasks.js
+++ b/src/store/modules/tasks.js
@@ -644,33 +644,26 @@ const actions = {
   assignSelectedTasks({ commit, state }, { personId, taskIds }) {
     const selectedTaskIds = taskIds || Array.from(state.selectedTasks.keys())
     return tasksApi.assignTasks(personId, selectedTaskIds).then(response => {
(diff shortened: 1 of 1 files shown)


NicoPennec:
This is a good quick-win. But I wonder if we shouldn't instead specify, at the time of selection, which tasks cannot be updated, in order to remove them from the Zou server call.

PR #1973, opened 2026-03-22
feat: Add support for sequence and episode entity types to the status automation modal and localization

mxatmx:
Pretty self explanatory. I am PR-ing a sister commit in the Zou repo.


mxatmx:
> We can also add the "Edit" entity type as well.

I'll get it done when I get a chance.

mxatmx:
@NicoPennec "edit" type added. Sorry for the delay.

PR #2001, opened 2026-04-30
feat: add timeout to authenticated api endpoint

mxatmx:
**Problem**
The authenticated API can hang indefinitely.

**Solution**
Adds a 10 second timeout to the ```/api/auth/authenticated``` endpoint.

### DIFF
--- a/src/lib/auth.js
+++ b/src/lib/auth.js
@@ -12,7 +12,7 @@ import {
 } from '@/store/mutation-types.js'
 
 let channel
-const AUTHENTICATED_REQUEST_TIMEOUT_MS = 10000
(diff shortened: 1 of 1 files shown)


NicoPennec:
Thank you for your improvement.

One minor drawback: 10 seconds might not be enough. A brief network interruption (VPN, mobile network, or cold restart) could cause the user to be disconnected.

Would a 20 or 30 second delay, or a single retry before the USER_LOGIN_FAIL error, be a good trade-off?

mxatmx:
> Would a 20 or 30 second delay, or a single retry before the USER_LOGIN_FAIL error, be a good trade-off?

@NicoPennec I suppose a 20 second delay could be fine. The goal here was just to get rid of this pesky issue. I'll make the change.

PR #2013, opened 2026-05-15
Fix annotation bugs in preview players

NicoPennec:
**Problem**

Multiple bugs in the preview annotation undo/redo mechanics: state shared across instances, side-effect pushes polluting the stack, stale `fabric.Object` refs after fullscreen exit, double additions payload, lost positions on `ActiveSelection` delete.

**Solution**
- per-instance undo stacks and silent-mode flag
- snapshot/restore around `addObject`/`deleteObject` in undo/redo
- `markRaw` on annotation save payloads
- resolve action targets by ID from the live canvas
- pass DOM elements to `fabric.Canvas` (avoid id collisions)
- drop redundant `addToAdditions` on 'remove' undo
- `discardActiveObject` before deleting an `ActiveSelection`
- add unit tests covering the undo/redo invariants

### DIFF
--- a/src/components/mixins/annotation.js
+++ b/src/components/mixins/annotation.js
@@ -1026,19 +1026,19 @@ export const annotationMixin = {
      */
     undoLastAction() {
       const action = this.doneActionStack.pop()
-      if (action && action.obj) {
-        if (action.type === 'add') {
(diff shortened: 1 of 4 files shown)


frankrousseau:
Included in the full refactoring

PR #2019, opened 2026-05-27
Fix subsequent annotations not saving from side panel

NicoPennec:
**Problem**
- When drawing annotations from the side panel preview (Shots, Concepts, Todos, AllTasks, TaskType…), only the first annotation is saved; subsequent ones are silently dropped.

**Solution**
- Fix subsequent annotations not saving from side panel

`TaskInfo`'s annotation handler was missed in ab8c3ae and never confirmed the save back to the player, leaving it stuck. Align it with the other consumers (`Task`, `Edit`, `Playlist`, `ViewPlaylistModal`).

### DIFF
--- a/src/components/sides/TaskInfo.vue
+++ b/src/components/sides/TaskInfo.vue
@@ -1079,17 +1079,22 @@ export default {
       this.currentPreviewDlPath = this.getOriginalDlPath()
     },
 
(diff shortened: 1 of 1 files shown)


frankrousseau:
Closed in favor of the refactoring

PR #2044, opened 2026-06-09
[players] Migrate annotation stack to fabric.js 7

frankrousseau:
## Problems

- Kitsu was pinned to `fabric@6.9.1` while fabric 7 is the current major (canvas 3, node 20+, jsdom 26).
- fabric 7 flips the default `originX/originY` to `'center'`: every Kitsu shape is created and serialised by its top-left corner, so new drawings and previously-saved annotations would render shifted by half a dimension.
- `fabricjs-psbrush` was still pinned to a personal fork branch instead of the upstream cgwire repo.
- Clicking an annotation marker landed on the previous frame about half the time (float floor vs the round used to draw the marker).
- Shift-scrub did not drive the comparison player in the playlist view.
- Drawing while navigating frames quickly could lose annotations: the canvas reloads asynchronously without cancellation, so a save could replace an annotation with a partially-loaded canvas, and legacy unrounded annotation times left duplicated same-frame entries of which only one ever displays.

## Solutions

- Bump `fabric` to `7.4.0` and repoint `fabricjs-psbrush` to `cgwire/fabricjs-psbrush` (peer range widened to `^6 || ^7` in cgwire/fabricjs-psbrush#2, prerequisite).
- Restore the v6 `left`/`top` default origin globally via `FabricObject.ownDefaults` at annotation module init, with a regression test.
- Compute the clicked annotation frame with `Math.round`, matching the marker position; bar clicks keep `floor`.
- Sync the comparison player from the shift-scrub handler, like every other navigation path.
- Cancel in-flight annotation loads on canvas clear (load token), merge canvas objects by id instead of replacing the annotation drawing on save, and merge duplicated same-frame entries at load time (`mergeAnnotationsByFrame`). Server-side cleanup of the legacy entries: cgwire/zou#1105.
- Gates: full unit suite (909 tests) and prod build green; runtime checklist validated by hand (draw + reload of every tool, pre-v7 annotation data, eraser mask, fullscreen/resize, playlist comparison, right-click).


frankrousseau:
Follow-up (non-blocking): the cgwire/fabric.js v5-snapshot fork is no longer referenced by Kitsu since the v6 migration — it can be archived once we confirm nothing else depends on it.

frankrousseau:
The annotation-loss fix in this branch has a server-side complement: cgwire/zou#1105 adds a `zou normalize-annotation-times` command that merges the legacy duplicated entries in base and snaps their times onto the player's frame grid. Until it is run, deleting a legacy-timed object can make it reappear on reload (the deletion targets the canonical time, not the stored legacy one).

PR #2054, opened 2026-06-15
[login] Fix OIDC/SAML SSO login labels and button cleanup

frankrousseau:
Follow-up to #2051 (OIDC SSO login button).

## Problems

- #2051 added the OIDC login keys to all non-English locales with the English text — and missed `de`/`hu` entirely — so non-English users see English and the keys look already translated (never flagged for translation).
- The SAML button's no-IdP-name fallback references `login.saml`, but the key is named `login_saml`: a missing key / broken fallback (latent, only hit when SAML is enabled without an IdP name).
- Both SSO buttons bind `is-loading` to `isLoginLoading` (the credential-login spinner), but they are `<a>` full-page redirects, so it never reflects their own click.

## Solutions

- Translate `login_with_oidc` / `oidc` in every locale by mirroring that locale's existing SAML strings (`saml_idp_name` → `oidc_idp_name`); add the missing keys to `de`/`hu`.
- Rename the fallback keys `login_saml` → `saml` and `login_oidc` → `oidc` so each provider is the pair `{login_with_X, X}`. This also fixes the SAML fallback, which already referenced `login.saml`.
- Drop the inert `is-loading` binding from both SSO buttons.
- `en.js` (source of truth) values unchanged.


NicoPennec:
`de.json` is missing

PR #2065, opened 2026-06-25
[entities] Add display toggle to show full task type names

mxatmx:
Add a "Show full task type names" option to the display settings on the shot, asset, sequence, episode and edit list pages. When enabled, each task type column auto-expands to fit its full name on one line instead of the fixed-width truncated (ellipsis) header. Off by default; persisted per entity type via the existing display_settings preference.


frankrousseau:
Thank you!

PR #2070, opened 2026-07-01
[shots] Add bulk shot generation to the Create shots modal

mxatmx:
Add a "Bulk" tab to the Create shots modal that generates a range of shot names from a start name, count and step (e.g. SH010, SH020, ...), previews them, and creates them in one pass via the existing CSV import. Adds a `generateBulkShotNames` string helper and a `bulkCreateShots` store action.


src/components/pages/Shots.vue
```diff
@@ -651,6 +653,18 @@ export default {
       this.newShot(shot).then(callback).catch(console.error)
     },
 
+    addShotsBulk(payload, callback) {
```
frankrousseau, line 656: Please don't use callback. Use only promises.

frankrousseau:
Two small guards worth adding to the bulk tab:

1. **TV show with no episode selected → empty Episode cell in the CSV.** `isBulkGenerateAllowed` only requires `selectedSequenceId`; on a TV show with `selectedEpisodeId` null, `episodeName ?? null` ends up as `""` in the generated CSV and the result depends on how zou handles an empty episode column. Adding `(!isTVShow || selectedEpisodeId)` to the condition closes it.

2. **Negative step is not blocked.** `min="1"` on the input doesn't prevent typing `-10`, and the `!bulk.step` guard only catches 0. Result: preview names like `SH-10`, `SH-20`, creatable if there is no collision. A `bulk.step < 1` check in the `bulkPreviewNames` guard is enough.

Minor, non-blocking: on import failure the modal only does `console.error` + spinner reset — with up to 500 shots at stake, surfacing an error message in the modal would help.

PR #2073, opened 2026-07-03
[shots] Fix Sequence Stats displaying empty on first load

### LINKED ISSUE #1900: Sequence Stats displays empty on first load
**Context**
Kitsu version: 0.20.109
Production type: Short

**Describe the bug**
My default landing is the Shots page.
Entering Sequence Stats page fails to display until manually refreshing.

**Screenshots**
Image

**Desktop (please complete the following information):**
 - OS Linux
 - Browser Chrome
 - Version 140

frankrousseau:
## Problems

- Opening Sequence Stats while shots were still loading (e.g. coming from the Shots landing page) computed stats on an empty shot cache and never recomputed, leaving the page empty until a manual refresh. Fixes #1900

## Solutions

- `loadShots` now queues callbacks made during an in-flight load (and returns a promise for await callers) and runs them once that load settles, instead of invoking them immediately on the empty cache

### DIFF
--- a/src/store/modules/shots.js
+++ b/src/store/modules/shots.js
@@ -83,6 +83,16 @@ const cache = {
   result: []
 }
 
+// Callbacks of loadShots calls made while another load was in flight.
+// They run once that load settles, so callers never see an empty cache.
+let pendingLoadShotsCallbacks = []
+
+const flushPendingLoadShotsCallbacks = err => {
+  const callbacks = pendingLoadShotsCallbacks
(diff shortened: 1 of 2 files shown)


frankrousseau:
Superseded: loadShots is now promise-only and concurrent callers share the in-flight promise (same pattern as the assets/edits fix) — implemented in #2092 (commit f4931f423) with a regression test. The callback-queue approach here no longer applies after the callback param removal.

PR #2089, opened 2026-07-03
[schedule] Fix "Invalid date" sent by the schedule Assign Tasks panel

frankrousseau:
## Problems

- With "display durations in hours" enabled, `getDatesFromStartDate`/`getDatesFromEndDate` skipped the business-days computation, leaving a null due date that the schedule Assign Tasks panel serialized as the literal string "Invalid date" (400 from the API). Fixes #1977

## Solutions

- Drop the display-preference guard: every caller passes the estimation in days (`minutesToDays` converts regardless of the flag), so the computation is valid for hours-displaying studios too — covered by a unit test

### DIFF
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -180,7 +180,10 @@ export const getDatesFromStartDate = (
   estimation,
   daysOff = []
 ) => {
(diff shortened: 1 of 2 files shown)


Magasul:
<3

PR #2112, opened 2026-07-06
[shots] Add per-sequence summary rows with collapse toggle

### LINKED ISSUE #1714: Summary of Sequence listed adjacent to it in Shot Tasks page
Would it be possible to have a summary in the shots tasks page for sequences - eg a sum of the time spent and frames aligned with the TIME & FRAMES columns 
eg 

                         TIME.   FRAMES.  <-- these line up with the already existing Time & Frames columns 
sequence01.      3.5         2046

Would also be useful to be able to expand / collapse the sequence and just display the summary (possibly with colour-coded sections, lined up under each task column - if one task isn't green, that task is the lowest task status present (eg if there is one wip and two CBB's, the WIP could be the lowest due to status priority).

frankrousseau:
## Problems
- The shots page gave no per-sequence overview: no aggregated time spent or frames aligned with the existing columns, and no way to collapse a sequence to a single summary line (fixes #1714)

## Solutions
- Extend the sequence header row of the shot list with totals aligned under the existing columns: time spent, estimation, frames and drawings sums for the sequence's non-canceled shots
- Show, under each task-type column, compact color-coded counts of shots per task status (status colors, tooltip with status name), for both sticked and regular columns
- Add a chevron next to the sequence name that collapses/expands the sequence's shot rows, keeping only the summary visible (local state, no persistence)
- Add a pure `computeGroupSummary` helper in `src/lib/stats.js` with unit tests, and `shots.collapse_sequence` / `shots.expand_sequence` locale keys with the video-game overlay overrides

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@fullcalendar/vue3": "6.1.21",
         "@google/model-viewer": "4.3.1",
         "@sentry/vue": "10.63.0",
+        "@tanstack/vue-virtual": "^3.13.31",
         "@unhead/vue": "3.1.7",
         "@vuepic/vue-datepicker": "14.0.0",
         "bowser": "2.14.1",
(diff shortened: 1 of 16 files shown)


frankrousseau:
Rebased on top of #2119 (list virtualization): the collapse now filters the virtualizer's flattened items instead of the template's v-for, and the summary row keeps the a11y keyboard handlers. Merge #2119 first — this PR's diff will then reduce to the feature commit only.

PR #2126, opened 2026-07-08
[lists] Make Name, Descriptions and rest of fields o nAssets list sortable

ivancea:
## Problem

Name, Description and other asset columns weren't sortable in the asset lists (Assets, Shots, Sequences, Episodes).

Also, clicking outside the dropdowns didn't close them, which allowed to open multiple dropdowns from multiple columns, feeling like the UI was broken.

## Solution

Make them sortable with a custom dropdown (Just the "Sort" option for now).

Detect outside clicks and close dropdowns.

### Notes

This is my first contribution in the UI, and as a user I only used the Assets list, so tell me if there's anything wrong, either functionally or technically!


frankrousseau:
Thank you!

PR #2128, opened 2026-07-11
[annotations] Delete fully erased annotations

ivancea:
## Problem

When erasing annotations, even after fully erasing them, they are still left as invisible selectables.

For example, given these annotations:
imagen

After erasing the 'B' letter:
imagen
imagen

## Solution

When an annotation is fully erased, remove it, so it doesn't add noise when selecting.

- Render erased annotations offscreen and inspect their alpha pixels.
- Delete objects only when no meaningful pixels remain.
- Keep partially erased objects and preserve undo/redo.
- Enable per-pixel hit testing so masked areas are not selectable.
- Cover full, partial, thin-stroke, persistence, and undo/redo cases with tests.

### DIFF
--- a/src/components/players/annotations/AnnotationCanvas.vue
+++ b/src/components/players/annotations/AnnotationCanvas.vue
@@ -124,6 +124,7 @@ const createFabric = () => {
     : new Canvas(canvasEl.value, {
         fireRightClick: true,
         enablePointerEvents: true,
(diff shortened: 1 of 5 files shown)


ivancea:
https://github.com/cgwire/kitsu/pull/2118 could affect this logic (Maybe?), so I'll review this after that one is merged. In case there's some weird interaction

PR #2135, opened 2026-07-13
[annotations] Preserve pressure stroke round caps

ivancea:
## Problem

The bounding-box around the annotation stroke wasn't accounting for the full stroke width, showing a "cut" stroke.

Before:
imagen

After:
imagen

## Solution

Remove a monkey-patch that was affecting this calculation.

I couldn't find if there was an (still existing) reason for it. I see it was when Kitsu was using Fabric 5. But not sure of the exact reason.

### DIFF
--- a/src/composables/players/annotation.js
+++ b/src/composables/players/annotation.js
@@ -52,15 +52,7 @@ Text.prototype._calculateCurrentDimensions = function () {
   )
 }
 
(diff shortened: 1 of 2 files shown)


frankrousseau:
Thank you, it was there because it was needed with Fabric 5. But Fabric 7 adds a point itself.

PR #2150, opened 2026-07-20
[assets] Support opening tasks in new tabs

ivancea:
## Problem

In the assets list, the tasks let you click on them to open the sidebar. However, you can't "Right click -> Open in a new tab" nor middle-click it to open it.
imagen

## Solution

This PR makes the task cells work as hrefs for right click and middle click, making it easy to open tasks in different tabs.
(Hide whitespace when reviewing!)

(This is just a PR-suggestion)


frankrousseau:
Thank you for the suggestion. We have to run some usage tests before integrating it.

PR #2166, opened 2026-07-30
[logs] Filter both log screens by date range, users and event type

nledez:
**Problem**
  - Activity logs and login logs only show one day at a time and offer a single user
  - That user filter runs in the browser over the page already loaded (1000 events, 100 logins), so it lies past the first page: picking an active person on a busy day can
  return nothing
  - Users cannot be searched by email, although `buildPeopleIndex` already supports it and is simply not wired into `PeopleField`
  - Activity logs expose none of the filters the API supports: production, event name, file events
  - An event line shows the action truncated to three letters, no production, no seconds, and only `project_id` / `task_id` are clickable

**Solution**
  - Both screens take a date range and several users at once, filtered server side, with search by first name, last name or email
  - Activity logs also filter by production, object, action and file events; the whole filter state lives in the query string so a URL can be shared mid-investigation
  - Object and action lists come from `/data/events/names`, so plugin events are offered like any other
  - The person filter is read from the route rather than from the widget model — the person map is filled asynchronously and would otherwise silently drop the filter on a cold
  page load
  - Event lines show the production, the full action name and timestamps with seconds; the detail panel links assets, shots, sequences, episodes, edits and persons (keys with no
  detail route, such as `comment_id`, stay unlinked rather than pointing at a dead URL)
  - CSV export of what is on screen, filters included
  - New `DateRangeField` and `MultiselectField` widgets; `PeopleField` now honours the `multiple` prop it already declared and gains an opt-in `search-email`


src/components/pages/logs/EventLogs.vue
```diff
@@ -220,44 +358,98 @@ const loadMoreEvents = async date => {
...
+  const escape = value => `"${String(value).replace(/"/g, '""')}"`
+  const csv = [headers, ...rows]
+    .map(row => row.map(escape).join(','))
+    .join('\n')
+  const link = document.createElement('a')
+  link.href = URL.createObjectURL(new Blob([csv], { type: 'text/csv' }))
+  link.download = `activity-logs-${formatSimpleDate(new Date())}.csv`
+  link.click()
+  URL.revokeObjectURL(link.href)
 }
```
NicoPennec, line 403: You can reuse the `csv.buildCsvFile` helper instead to rewrite your own export method.

src/components/pages/logs/EventLogs.vue
```diff
@@ -106,91 +152,184 @@ import {
...
+    .filter(Boolean)
+}
+
+// Seconds matter here: correlating events logged within the same minute is
+// the whole point of an audit trail.
+const formatEventDate = date => {
+  const localDate = moment.tz(date, 'UTC').tz(timezone.value)
+  const timeFormat = use12HourClock.value ? 'h:mm:ss A' : 'HH:mm:ss'
+  return `${localDate.format(dateFormat.value)} ${localDate.format(timeFormat)}`
 }
```
NicoPennec: The `formatTimeOfDay` helper method seems to do the same things.

src/components/widgets/DateRangeField.vue
NicoPennec, line 1: Maybe just include a `range` option in `DateField.vue` instead of creating a fork.

src/components/pages/logs/EventLogs.vue
```diff
@@ -1,34 +1,81 @@
...
+
+    <div class="flexrow filters mt1">
+      <label class="flexrow-item checkbox-label">
+        <input
+          type="checkbox"
+          v-model="onlyFiles"
+          @change="onOnlyFilesChange"
+        />
+        {{ $t('logs.only_files') }}
+      </label>
```
NicoPennec: You can reuse the `Checkbox.vue` component.

src/locales/en.js
```diff
@@ -721,9 +721,19 @@ export default {
   },
 
   logs: {
+    action_label: 'Action',
+    action_placeholder: 'All actions',
     current_date_label: 'Show logs for',
```
NicoPennec: The `current_date_label ` is no longer used in this PR, so it can be removed.

src/locales/fr.json
NicoPennec, line 1: The new i18n workflow now involves translating all locales.

NicoPennec (approved):

PR #2169, opened 2026-08-03
[tasks] Show the comment box to someone who was mentioned

tinnhcgi:
### Problem

Mentioning someone in a task sends them a notification and an email, they open the task — and there is no comment box. `isCommentingAllowed` covers assignees, clients, department supervisors and managers only, so the mention brings them in and the UI gives them no way to answer.

### What this changes

A mention counts as admission to the conversation. `isCommentingAllowed` and `is-replyable` also accept someone named in a comment of that task, directly or through one of their departments. The comment payload already carries `mentions` and `department_mentions`, so this needs no extra request.

Someone who is *only* there because they were mentioned should not move the task, so `AddComment` takes `isStatusLocked`: the status selector is hidden and the task's current status is sent. This is purely about not offering an action that would be refused — the API enforces the rule.

### Requires the API change

The permission check lives in Zou: cgwire/zou#1184. **Please do not merge this one alone** — without it the comment box appears and the API rejects the post.

### Status

Running in production since 2026-08-03 alongside the Zou patch, on 1.0.55 / Zou 1.0.62. Mentioned artists can answer, they see no status selector, and preview upload is still refused. `prettier --check` passes on the three files.

### DIFF
--- a/src/components/pages/Task.vue
+++ b/src/components/pages/Task.vue
@@ -317,6 +317,7 @@
                 :is-loading="loading.addComment"
                 :is-movie="isMovie"
                 :is-picture="isPicture"
+                :is-status-locked="isMentionedOnly"
                 :team="currentTeam"
                 :task-types="currentTaskTypes"
                 :task="task"
@@ -365,6 +366,7 @@
                     :is-replyable="
                       (user && user.id === comment.person?.id) ||
                       isAssigned ||
+                      isMentioned ||
                       isDepartmentSupervisor ||
                       isCurrentUserManager
                     "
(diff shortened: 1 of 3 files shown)


tinnhcgi:
Updated: `isMentioned` now also walks `comment.replies`, not just the comment itself.

Replies carry their own `mentions` / `department_mentions`, and that is where people usually get named once a conversation is going — the first version missed those. The API side in cgwire/zou#1184 was updated to match, so the two stay consistent.

`prettier --check` still passes on the three files.

frankrousseau:
Thank you @tinnhcgi for your contributions!

PR #2182, opened 2026-08-11
[preview] Add image background color picker

ivancea:
**Problem**
For images (previews) with transparent background, the by-default black background will sometimes hide details, or have a bad contrast.

**Solution**
Add a selectable background color on the preview:
imagen
imagen

Same in the fullscreen preview:
imagen

Background icon, placement in the tools and such can be changed.

Not tested with other kinds of assets; I lack knowledge on those areas of Kitsu. Would need guidance here


frankrousseau:
Thank you!

PR #2194, opened 2026-09-01
[filters] Match empty metadata values

ivancea:
**Problem**

- Filters such as [Invoice]=[] cannot find assets with blank metadata.
- Blank metadata can be stored as a missing key, null, an empty string, or an empty tag list.

**Solution**

- Match each datatype's blank representation when filtering with [].
- Keep non-empty metadata filters unchanged.
- This cannot distinguish never-set from cleared metadata because both appear as blank values in Kitsu.

### DIFF
--- a/src/lib/filtering.js
+++ b/src/lib/filtering.js
@@ -97,7 +97,9 @@ const applyFiltersFunctions = {
       (filter.descriptor.entity_type === 'Task'
         ? undefined
         : entry.entity_data?.[filter.descriptor.field_name])
-    if (dataValue !== undefined && dataValue !== null && filter.values) {
+    if ((dataValue === undefined || dataValue === null) && filter.values) {
+      isOk = filter.values.includes('')
+    } else if (filter.values) {
       if (typeof dataValue === 'string') dataValue = dataValue.toLowerCase()
 
       // Checklist case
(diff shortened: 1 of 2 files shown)


src/lib/filtering.js
```diff
@@ -97,7 +97,9 @@ const applyFiltersFunctions = {
       (filter.descriptor.entity_type === 'Task'
         ? undefined
         : entry.entity_data?.[filter.descriptor.field_name])
-    if ((dataValue || dataValue === 0) && filter.values) {
+    if ((dataValue === undefined || dataValue === null) && filter.values) {
```
NicoPennec: Folding `''` into the blank check also covers number descriptors, where it currently falls through to `parseFloat('')` and never matches.

```suggestion
    const isBlank =
      dataValue === undefined || dataValue === null || dataValue === ''
    if (isBlank && filter.values) {
```

NicoPennec:
Thanks @ivancea! This looks good to me overall.

The only thing I would change is the number case: 
`[Budget]=[]` still misses cleared number metadata, since `''` skips the new missing/null branch and `parseFloat('')` is `NaN`, so normalizing blank once before the `data_type` dispatch would cover every type.

